PR #12, opened 2020-11-03
[WIP] Add some ability to disable tictoc output

DLWoodruff:
I added a function to disable tictoc output, but it won't yet disable the output created by __init__; I could live with that for
now, but maybe someone has a good idea for how to disable that as well (and still get a really good tictoc starting point, which
is also important)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,10 @@ mpi-sppy
 
 Optimization under uncertainty for Pyomo models.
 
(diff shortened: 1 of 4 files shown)


bknueven:
We could direct the mpisppy.__init__ timer output to devnull--then we'd never see it, but that's when the clock would start.

bknueven (approved):
No issues here.

I think eventually we should wrap TicTocTimer for our specific use-case (to turn in off, not requiring `delta=False`, etc.)

PR #13, opened 2020-11-04
Move examples

bknueven:
- Also fixes acopf3 example for upstream Egret changes, lots of other small adjustments to make existing code work.


DLWoodruff (approved):
Good change.

PR #17, opened 2020-11-15
Track pyomo tests

DLWoodruff:
Add an action to test against Pyomo master every night. I also think I correctly added a badge for this.


bknueven (approved):
Looks good. Let's merge and make sure everything works as expected.

PR #22, opened 2020-12-03
Using unconditional probability

bknueven:


jeanpaulwatson:
Developed interactively with Ben - all looks good.

PR #26, opened 2020-12-16
Docfix

dtmildebrath:
I started to write a sort-of "tutorial" so that users can see a hello-world-like example in the documentation. This doesn't need to be merged now--I'm just looking for feedback on whether this endeavor is good or useful. The writeup is in `examples.rst`. The documentation can be tested with
``` 
make doctest 
```
and compiled with
```
make html
 ```

Other minor changes:
* Added a utility function in the `ExtensiveForm` class to retrieve the objective value.
* Added a page for references in the documentation. 
* Corrected typos.

### CODE REVIEW

**Review by DLWoodruff — APPROVED** [review]
My bad. Thanks for fixing it.

**Review by DLWoodruff — COMMENTED** [review]
This is a good way to go. We can easily add a github action for the doctests.

File: mpisppy/opt/ef.py [hunk 1]
```diff
@@ -88,6 +88,23 @@ def solve_extensive_form(self, solver_options=None, tee=False):
...
+                Objective value.
+
+        Raises:
+            ValueError:
+                If optimal objective value could not be retrieved.
+        """
+        try:
+            obj_val = pyo.value(self.ef.EF_Obj)
+        except Exception as e:
+            raise ValueError("Could not extract EF objective value with error: {str(e)}")
```
bknueven, line 105: Missing `f` at the beginning of this string

dtmildebrath:
I had to do a little modifying to the actual codebase in order to make it easier for casual users to actually extract the solutions from the models (I think there's more work to do in this direction, but that's for another PR).

DLWoodruff (approved):
These are all very good changes. We should merge this right away, but at our next meeting (if not sooner) let's put in a github action to run the doc tests.

DLWoodruff (approved):
This is PR is getting to be pretty big, but it is, at least, pretty good. I say we merge it.

PR #28, opened 2020-12-17
More numpythonic Compute_Xbar

bknueven:
Uses a bit of extra memory to move a several operations into C. This should obviously be carefully checked and verified.

Also renames `node_concats` to `global_concats` in Compute_Xbar.

### DIFF
--- a/mpisppy/phbase.py
+++ b/mpisppy/phbase.py
@@ -150,9 +150,9 @@ class PHBase(mpisppy.spbase.SPBase):
         nodenames = list() # to transmit to comms
         # This "onlyreduce" thing is done for the sake of asynchronous only
         local_concats = {"OnlyReduce": dict()} # keys are tree node names
-        node_concats =  {"OnlyReduce": dict()} # values are concat of xbar 
-                                               # and xsqbar
-        
+        global_concats =  {"OnlyReduce": dict()} # values are concat of xbar
+                                                 # and xsqbar
+
         # we need to accumulate all local contributions before the reduce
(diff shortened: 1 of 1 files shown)


DLWoodruff (commented):
Take a look at my comment on the issue and see what you think. I think precomputing the coefficients makes sense and the way I want to do it will obviously help the differing Vars use-case.

mpisppy/phbase.py
```diff
@@ -164,25 +164,29 @@ def Compute_Xbar(self, verbose=False, synchronizer=None):
...
-                        (s.PySP_prob / node.uncond_prob) * v._value * v._value
+                nlen = nlens[ndn]
+
+                xbars = local_concats["OnlyReduce"][ndn][:nlen]
+                xsqbars = local_concats["OnlyReduce"][ndn][nlen:]
+
+                nonants_array = np.fromiter( (v._value for v in node.nonant_vardata_list),
+                                             dtype='d', count=nlen )
+                xbars += (s.PySP_prob / node.uncond_prob) * nonants_array
+                xsqbars += (s.PySP_prob / node.uncond_prob) * nonants_array**2
```
bknueven, line 182: The way this is re-factored you could drop in `s._PySP_Prob_Mask[ndn]`/`s._PySP_Prob_Coeff[ndn]` (as a numpy array) for `(s.PySP_prob / node.uncond_prob)`.

In the typical case (same variable probabilities), `s._PySP_Prob_Coeff[ndn]` could just be set to `(s.PySP_prob / node.uncond_prob)` (and not use an array).

Issue #9, opened 2020-11-02
Pull in find_active_objective from PySP


bknueven:
This would weaken mpi-sppy's dependence on PySP to a single module: `mpisppy.utils.pysp_model`.

bknueven:
Closed by #39.

PR #39, opened 2021-01-07
Findobj

DLWoodruff:
Dropping dependence on PySP  by adding my version of an objective function finder. This fixes issue #9

### DIFF
--- a/mpisppy/utils/sputils.py
+++ b/mpisppy/utils/sputils.py
@@ -748,17 +748,15 @@ def reenable_tictoc_output():
     
 def find_active_objective(pyomomodel):
     # return the only active objective or raise and error
-    cnt = 0
-    for obj in pyomomodel.component_data_objects(Objective, active=True,
-                                                 descend_into=True):
-        if cnt == 0:
-            retval = obj
-        cnt += 1
-    if cnt > 1:
-        raise RuntimeError(f"Multiple Objectives found for {pyomomodel.name}")
-    else:
-        return retval
-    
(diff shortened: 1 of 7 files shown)


mpisppy/utils/sputils.py
```diff
@@ -744,6 +744,20 @@ def reenable_tictoc_output():
     # Primarily to re-enable after a disable
     tt_timer._ostream.close()
     tt_timer._ostream = sys.stdout
+
+    
+def find_active_objective(pyomomodel):
```
bknueven, line 749: I think we could simplify the implementation of this function:

retval = None
for obj in (...):
    if retval is None:
        retval = obj
    else:
        raise RuntimeError(...)
return retval

bknueven (approved):
Looks good

mpisppy/utils/sputils.py
bknueven: Maybe we should also be raising an exception if there's *no* objective..

jsiirola: This seems excessively complicated.  What do you think of the following:
```python
 def find_active_objective(pyomomodel):
    obj = list(pyomomodel.component_data_objects(
        Objective, active=True, descend_into=True)
    if len(obj) != 1:
        raise RuntimeError("Could not identify exactly one active "
                           "Objective for model '%s' (found %d objectives)"
                           % (pyomomodel.name, len(obj)))
    return obj[0]
```

DLWoodruff: It pays to have friends. I will make the change.

PR #40, opened 2021-01-07
Refactoring ph_term

bknueven:


DLWoodruff (approved):
The is better, clearer and also correct.

PR #41, opened 2021-01-10
NOW: minor error message improvement; WAS (but is no longer): xhatbase now calls the post_iter0 for an extension *on* *its* opt object

DLWoodruff:
This PR is being superseded by a branch Ben started that allows for callbacks after every subproblem solve. I am going merge it with the (unrelated) improvement to an error message.

Old Text:
I happen to need to have xhatters call the post_iter0 on the extension on the opt object that they own (never mind that the
xhatter itself might be an extension). 

As luck would have it, it turns out that post_iter0 is the only callback that was easy.  I'm not quite sure how, or even whether, the other extension calls should be supported.  (Iter 0 is very special in our implementation of PH, so the post-iter0 extension
call is sometimes the only one needed).

### DIFF
--- a/mpisppy/extensions/xhatbase.py
+++ b/mpisppy/extensions/xhatbase.py
@@ -215,10 +215,6 @@ class XhatBase(mpisppy.extensions.extension.PHExtension):
     def post_iter0(self):
         # the base class needs this
         self.comms = self.opt.comms
-        # As long as were are here, we might as well check
-        # to see if the underlying opt object itself has an extension.
-        if hasattr(self.opt, "extobject") and self.opt.extobject is not None:
-            self.opt.extobject.post_iter0()
         
 """
 May 2020, DLW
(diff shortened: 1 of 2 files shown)


dtmildebrath:
I'm tired right now so maybe I'm missing something, but couldn't this lead to an extension object recursively calling its own `post_iter0` method infinitely? The extension has a pointer to the opt object, and the opt object has a pointer to the extension. So it seems like this would repeat ad infinitum? Please set me straight

DLWoodruff:
dtmildebrath raises a good point. We would want to add code to check to see that the extension's opt object did not have the extension object itself as an extension object.  Needless to say, that's not the use case.  My particular use case is that 
modifications to the Pyomo model need to be done after iter0. The hub and any xhatter cylinders need to make these modifications using a post_iter0 extension.  For pretty good reasons, the xhatters are implemented as extensions.

bknueven:
Maybe out-of-scope for this PR, but it suggests we should be carefully thinking about how extensions interact with cylinders.

For example, it isn't uncommon to need to modify a pyomo model after solve (say to add constraints). Perhaps we could generalize (some of) the concepts in the PH extensions to "post first-solve" and "after every solve". Both could be useful for Lagrangian and Xhatter extensions.

PR #43, opened 2021-01-11
Proximal term approximation

bknueven:
Adds the ability to linearize proximal terms for PH, which connections in baseparsers and vanilla.

Option interface:
* `PHoptions['linearize_proximal_terms']` : flag for if to use an approximation of the proximal terms for all variables. If set, ignores the value in `PHoptions['linearize_binary_proximal_terms']`.
* `PHoptions['initial_proximal_cut_count']` : number of pieces to use in the initial approximation of the proximal term (default `2`)
* `PHoptions['proximal_linearization_tolerance']` : After each PH iteration, we measure the distance between the linearization and xhat. If it's greater than this tolerance, we add a cut by projecting xhat onto y == x^2 (default `1.e-1`).


DLWoodruff:
initial response:
'initial_proximal_term_cuts' should be 'initial_proximal_cut_count'

DLWoodruff:
A few thoughts (maybe this is not the right place to work through my understanding):
           for prox_approx_manager in s._xsqvar_prox_approx.values():
                prox_approx_manager.check_tol_add_cut(tol, persistent_solver)
caught my eye. I guess there are multiple approx objects because there is one for each variable, which makes sense. It also makes sense that the baseparser and vanilla objects would be set up to support doing all Vars or none. But I assume that a clever programmer could decide to have the linear approximation for some Vars and use the quadratic version for others. Is all that correct?

bknueven:
Yes -- in addition you could use different tolerances / cut initializations for each nonant variable, though none of these are exposed in the current interface.

dtmildebrath:
Out of curiosity, are we seeing bottlenecks/performance degradation due to the quadratic solve for any particular class of problems? Or is this just a general improvement

DLWoodruff:
Ben: When you use the word "interface," I think you are referring to what is supported in baseparsers and vanilla, right?

David M: We saw major speed-ups (factor of 5 or more) in PySP using this. Historically, the QMIP solvers were maybe an order of magnitude slower that the MIP solvers (I have not checked lately).   You might not expect that, but that is what we were seeing.

bknueven:
@DLWoodruff by "interface" here I mean the new `PHoptions`; e.g., if we wanted more granular control over individual variables proximal term approximation we would need to poll a callback or something similar.

bknueven:
@dtmildebrath I would add is that the method here is quite a bit more "sloopy" than what a QMIP solver would do under-the-hood. We only add a single cut for each variable per PH iteration, allowing the cuts to accumulate over time, as opposed to *solving* the MIQP at each iteration. (Perhaps we should be setting a cut-maximum per variable, though that can be implicitly controlled with the tolerance.) The idea being that the approximation doesn't need to be great to start out, and becomes better and better at the point PH is zeroing in on.

It also enables PH to be used with open-source MIP solvers.

PR #50, opened 2021-01-18
Adding global_toc

bknueven:
This PR adds a wrapper around `mpisppy.tt_timer.toc` called `global_toc`. `mpisppy.global_toc` takes two arguments, a message and an optional conditional. If the conditional is `False`, nothing will be printed, otherwise `mpisppy.tt_timer.toc` is called with `delta=False`. By default the conditional is only set to `True` on global rank 0 and `False` on all other ranks.


DLWoodruff (approved):
Much cleaner.

PR #53, opened 2021-01-21
Secretmenu

DLWoodruff:
I am adding a little documentation, but it is not yet doctested. I still think we should add it.


bknueven (approved):
Thanks for adding some documentation for the proximal term approximation.

PR #67, opened 2021-02-18
Issue 60

### LINKED ISSUE #60: Scenario subproblem namespacing
We should attach all the objects we currently attach to scenario subproblems to a single block on those subproblems, perhaps named `mpisppy`.

This will ensure we have exactly one name clash to worry about, and the internal `mpisppy` names can be less verbose.

bknueven:
~~TODO: I didn't dig too deeply into various files to see if we've gotten everything.~~

~~There's also the associated issue of renaming things like "_PySPnode_list" and "PySP_prob".~~ (This should be in its own PR as its very disruptive.)

Closes #60.


DLWoodruff:
Right. The next thing to do is grep for PySP and rename everywhere except the few places where we really mean it.

dtmildebrath:
`_look_and_leap` is beautiful!

dtmildebrath (approved):

PR #69, opened 2021-02-19
Replace cb_data with scenario_creator_kwargs

### LINKED ISSUE #66: Change scenario_creator function handle
...to pass in a scenario name and then arbitrary kwargs.

dtmildebrath:
Replaced the old `cb_data` interface with a more Pythonic `scenario_creator_kwargs` option. This closes #66.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,7 +69,7 @@ instance/
 .scrapy
 
 # Sphinx documentation
-docs/_build/
+doc/_build/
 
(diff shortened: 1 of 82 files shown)


DLWoodruff (approved):
It is pretty hard to check all these files. I think we need to merge and deal with whatever comes up. It is a valuable change.

bknueven (approved):

PR #74, opened 2021-02-26
norm rho updater

bknueven:
Adds a norm rho updater extension and converger. Makes a few changes to the converger API. More-or-less a re-implementation of the adaptive_rho_converger PySP plugin, with slightly different defaults and more settable parameters.

Example use of rho updater is added to farmer_cylinders.


DLWoodruff (approved):
As discussed in the meeting with David M, this looks good. I see that you sneaked in a little improvement to update_W as well.

PR #77, opened 2021-03-04
Renaming PySP_prob and _PySPnode_list

bknueven:


DLWoodruff (approved):
I would like to merge these changes. We have some testing issues (mainly with run_all.py) related to changes in Pyomo that we will have to endure until Pyomo 6.0 is released (which should happen in about a month).

PR #83, opened 2021-03-11
Pyomo6 mod LShaped

### LINKED ISSUE #19: tictoc timer
In preparation for Pyomo 6, we need to write our own timer. For one thing, we need to be able to have it go silent.

### LINKED ISSUE #82: pysp import
In preparation for Pyomo 6, we need to import pysp as its own package.

bknueven:
Closes #19 and closes #82.

Addresses import issues around TicTocTimer and PySP.

### DIFF
--- a/mpisppy/__init__.py
+++ b/mpisppy/__init__.py
@@ -1,18 +1,26 @@
 # Copyright 2020 by B. Knueven, D. Mildebrath, C. Muir, J-P Watson, and D.L. Woodruff
 # This software is distributed under the 3-clause BSD License.
-import os
-from pyutilib.misc.timing import TicTocTimer
+
+import pyomo as _pyo
+
+pyomo6 = (int(_pyo.__version__[0]) >= 6)
+
+if pyomo6:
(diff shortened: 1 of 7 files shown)


mpisppy/__init__.py
```diff
@@ -1,18 +1,26 @@
...
+pyomo6 = (int(_pyo.__version__[0]) >= 6)
+
+if pyomo6:
+    from pyomo.common.timing import TicTocTimer as _TTT
+else:
+    from pyutilib.misc.timing import TicTocTimer as _TTT
+
 try:
-    import mpi4py.MPI as mpi
+    import mpi4py.MPI as _mpi
```
dtmildebrath, line 14: Not particularly important but any reason we prepended `mpi` with an underscore?

bknueven: Only in case somebody does `from mpisppy import *`. The names with leading underscores will *not* be imported (and they're local here anyhow).

DLWoodruff (approved):
Looks good.

PR #87, opened 2021-03-18
fix lack of W bug in APH

DLWoodruff:


jeanpaulwatson (approved):
I blame Python for not telling us that results of particular computations were thrown on the floor.

PR #98, opened 2021-04-27
[WIP] bundle info on scenarios to support aph

DLWoodruff:
subproblem_creation now adds bundle info to scenario data and aph uses it
Merge this only after the phbase refactor has merged (there will be minor conflicts)


bknueven:
Replaced by #106

PR #100, opened 2021-04-30
Quadratic term expansion for APH

jeanpaulwatson:
Expanding the square in the APH proximal term, to (partially) mitigate cost of canonical expression computation in Pyomo.


DLWoodruff (approved):
Looks good.

PR #101, opened 2021-05-04
Schur-Complement Interface

michaelbynum:
This PR adds an interface to solve continuous problems with Schur-Complement decomposition.

I also added a method to `ExtensiveForm` to override `gather_var_values_to_rank0` from `SPBase` so that `report_var_values_at_rank0` works correctly for `ExtensiveForm`.


mpisppy/opt/sc.py
```diff
@@ -0,0 +1,121 @@
...
+
+        models = list(local_scenario_models.values())
+        ref_model = models[0]
+        models = models[1:]
+        self.nonant_vars = list()
+
+        for node in ref_model._mpisppy_node_list:
+            ndn = node.name
+            for ndx, v in enumerate(node.nonant_vardata_list):
+                self.nonant_vars.append((ndn, ndx))
```
bknueven, line 48: `self.nonant_vars = list(ref_model._mpisppy_data.nonant_indices.keys())` would accomplish the same thing.

mpisppy/opt/sc.py
```diff
@@ -0,0 +1,121 @@
...
+
+        active_obj = _get_active_objective(m)
+        active_obj.deactivate()
+        m._mpispypy_weighted_obj = pyo.Objective(expr=m._mpisppy_probability * active_obj.expr, sense=active_obj.sense)
+
+        nonant_vars = dict()
+        for node in m._mpisppy_node_list:
+            ndn = node.name
+            for ndx, v in enumerate(node.nonant_vardata_list):
+                nonant_vars[ndn, ndx] = v
```
bknueven, line 69: Similarly, this same dictionary is available at `m._mpisppy_data.nonant_indices`.

mpisppy/opt/sc.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def build_model_for_scenario(self,
+                                 scenario_identifier: str) -> Tuple[_BlockData, Dict[Any, _GeneralVarData]]:
+        m = self.local_scenario_models[scenario_identifier]
+
+        _assert_continuous(m)
+
+        active_obj = _get_active_objective(m)
+        active_obj.deactivate()
+        m._mpispypy_weighted_obj = pyo.Objective(expr=m._mpisppy_probability * active_obj.expr, sense=active_obj.sense)
```
bknueven, line 63: It's a little more consistent with the mpi-sppy style if you add this objective to the block `m._mpisppy_model`.

mpisppy/opt/sc.py
```diff
@@ -0,0 +1,121 @@
...
+            raise RuntimeError(f'Variable {v} in block {m} is not continuous; The Schur-Complement method only supports continuous problems.')
+
+
+def _get_active_objective(m: _BlockData):
+    res = None
+    for obj in m.component_data_objects(pyo.Objective, descend_into=True, active=True):
+        if res is not None:
+            raise RuntimeError(f'Found multiple active objectives on {m}')
+        res = obj
+    return res
```
bknueven, line 29: This function is a repeat of `mpisppy.utils.sputils.find_active_objective`.

mpisppy/opt/sc.py
```diff
@@ -0,0 +1,121 @@
...
+                 mpicomm=None,
+                 model_name=None,
+                 suppress_warnings=False):
+        super(SchurComplement, self).__init__(options=options,
+                                              all_scenario_names=all_scenario_names,
+                                              scenario_creator=scenario_creator,
+                                              scenario_creator_kwargs=scenario_creator_kwargs,
+                                              all_nodenames=all_nodenames,
+                                              mpicomm=mpicomm)
+
```
bknueven, line 93: We should probably check two flags set by `SPBase.__init__`: `self.bundling` (I think should be `False`) and `self.multistage` (probably also should be `False`).

bknueven (commented):
Overall this looks like an excellent contribution, though I haven't tested it myself. I put in a few nits that would more deeply integrate this with existing helpers and attributes in mpi-sppy, but none are really blockers.

@DLWoodruff should we add examples/farmer/schur_complement.py to runall? Do we think anything here needs to wait for #94?

michaelbynum:
@bknueven Thanks for the quick review! I'll make these changes shortly.

mpisppy/opt/sc.py
michaelbynum: This is very helpful. Thanks.

mpisppy/opt/sc.py
michaelbynum: Thanks!

mpisppy/opt/sc.py
michaelbynum: Will do.

mpisppy/opt/sc.py
michaelbynum: Thanks!

mpisppy/opt/sc.py
michaelbynum: I was actually hoping this would work for multistage problems, but I have not tested it yet...

mpisppy/opt/sc.py
```diff
@@ -58,16 +45,12 @@ def build_model_for_scenario(self,
...
+        active_obj = find_active_objective(m)
         active_obj.deactivate()
-        m._mpispypy_weighted_obj = pyo.Objective(expr=m._mpisppy_probability * active_obj.expr, sense=active_obj.sense)
-
-        nonant_vars = dict()
-        for node in m._mpisppy_node_list:
-            ndn = node.name
-            for ndx, v in enumerate(node.nonant_vardata_list):
-                nonant_vars[ndn, ndx] = v
+        m._mpisppy_model = pyo.Block()
```
bknueven, line 50: Final nit; I wasn't clear before. `SPBase.__init__` already attaches this block, you just have to use it.
https://github.com/Pyomo/mpi-sppy/blob/76132b5dc03fae54b2c305e7dff9e43d92b092aa/mpisppy/spbase.py#L477-L482

DLWoodruff:
This looks like a great starting point. I accomplishes the goal of having SC use mpi-sppy scenario creator functions! There is a PR that needs to merge first (for one thing, the "gather_vars..." function moves in that PR.)

mpisppy/opt/sc.py
michaelbynum: Thanks again. Fixed!

michaelbynum:
> This looks like a great starting point. I accomplishes the goal of having SC use mpi-sppy scenario creator functions! There is a PR that needs to merge first (for one thing, the "gather_vars..." function moves in that PR.)

Understood.

DLWoodruff:
We merged the PRs that were blocking this, but then we modified report_var_values_at_rank0 with the intention of making
it work with the EF. At the time of this writing, report_var_values_at_rank0 is not yet working with the EF (I merged the changes anyway, because no functionality was lost and that makes it easier to test and improve).

michaelbynum:
Thanks @DLWoodruff. I removed my changes to the EF.

bknueven (approved):

michaelbynum:
I do not have permission to merge this. Is it waiting on a second review, or does someone just need to hit merge?

PR #102, opened 2021-05-04
only run GHA to publish to pypi for tagged commits

michaelbynum:
I believe this PR will ensure the "publish-to-test-pypi" tests only get run for tagged commits. This allows GitHub Actions to be enabled on forks.


bknueven (approved):
The existing behavior is probably too paranoid, and this test will always fail on forks.

PR #111, opened 2021-05-06
Adding presolve callback for SPOpt objects

bknueven:
...to mirror the existing `post_solve` callback.


DLWoodruff (commented):
This is obviously a good idea; however, both for this callback and for the post_solve callback, I think we need to
 Change the name from PH_extensions to something like extensions (or SPOpt_extensions?)

bknueven:
I think it should just be called `extensions`.

For standardization should *all* instances of `PH_extensions` in `SPBase`-derived classes be renamed `extensions`?

DLWoodruff:
I'm not sure about calling everything "extensions" because I think the constructor should set the extensions attribute to None and if different levels allow different extensions, they probably need different names.

bknueven:
The issue is our current extensions hierarchy doesn't make this easy, because a `PHExtension` is also an instance of `Extension`. So then the code in `SPOpt.solve_one` would need to know about both, if they have different names.

DLWoodruff:
I'm going to move this conversation to slack.

PR #112, opened 2021-05-06
Solution writers

bknueven:


DLWoodruff:
Change the default on restore_nonants to False? Maybe we should change it to default None. Spokes want it to be False, and a final callback wants it to be False, but if it were used as an extension (not post_everything) it needs to be True.  So if we are going by popularity, it would be False, but then if anyone ever uses it as a callback, they will probably get burned in silent, but deadly, ways.

DLWoodruff (approved):
I am clicking "Approve" but I would be happy to wait until Sunday for JP or David M to weigh in, but on Sunday morning, I would like to merge this.

bknueven:
My reasoning for putting the writers on `SPBase` was two fold:
1. `SPBase`, through the new attributes `tree_solution_available` and `first_stage_solution_available`, "knows" if it's valid to write the solution, and the writers themselves act on attributes of this class (namely `local_scenarios`).
2. If PH is run with an xhat extension by itself, or LShaped by itself, it seems most natural to have the writers on the PH/LShaped class

That said, I'm open to persuasion.

PR #113, opened 2021-05-11
Xhatnonants

DLWoodruff:
npy writers for xhat nonants
This is "tested" in run_all. At least the ef writer will be in a unit tests for MMW very soon.

### DIFF
--- a/examples/farmer/farmer_cylinders.py
+++ b/examples/farmer/farmer_cylinders.py
@@ -120,7 +120,8 @@ def main():
         sputils.write_spin_the_wheel_first_stage_solution(spcomm,
                                                           opt_dict,
                                                           'farmer_cyl_nonants.spy',
-                                                          first_stage_solution_writer=sputils.first_stage_nonant_npy)
+                                                          first_stage_solution_writer=\
+                                                          sputils.first_stage_nonant_npy_serializer)
         sputils.write_spin_the_wheel_tree_solution(spcomm, opt_dict, 'farmer_full_solution')
 
 if __name__ == "__main__":
--- a/examples/farmer/farmer_cylinders.py
+++ b/examples/farmer/farmer_cylinders.py
@@ -117,6 +117,10 @@ def main():
 
     if write_solution:
         sputils.write_spin_the_wheel_first_stage_solution(spcomm, opt_dict, 'farmer_plant.csv')
+        sputils.write_spin_the_wheel_first_stage_solution(spcomm,
+                                                          opt_dict,
+                                                          'farmer_cyl_nonants.spy',
+                                                          first_stage_solution_writer=sputils.first_stage_nonant_npy)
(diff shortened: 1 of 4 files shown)


examples/farmer/farmer_cylinders.py
```diff
@@ -117,6 +117,10 @@ def main():
 
     if write_solution:
         sputils.write_spin_the_wheel_first_stage_solution(spcomm, opt_dict, 'farmer_plant.csv')
+        sputils.write_spin_the_wheel_first_stage_solution(spcomm,
+                                                          opt_dict,
+                                                          'farmer_cyl_nonants.spy',
+                                                          first_stage_solution_writer=sputils.first_stage_nonant_npy)
```
bknueven, line 123: We should probably rename the kwarg `first_stage_solution_writer` to `first_stage_nonant_writer` for this function -- and maybe the function itself should be renamed `write_spin_the_wheel_nonant_solution`.

mpisppy/spbase.py
```diff
@@ -591,7 +591,7 @@ def report_var_values_at_rank0(self, header="", print_zero_prob_values=False):
                 print("")
 
     def write_first_stage_solution(self, file_name,
-            first_stage_solution_writer=sputils.first_stage_solution_writer):
+            first_stage_solution_writer=sputils.first_stage_nonant_writer):
```
bknueven, line 594: Similar here, we should rename the kwarg `first_stage_solution_writer` to `first_stage_nonant_writer`, and maybe rename this method `write_nonant_solution`.

mpisppy/utils/sputils.py
```diff
@@ -130,7 +130,14 @@ def spin_the_wheel(hub_dict, list_of_spoke_dict, comm_world=None):
 
     return spcomm, opt_dict
 
-def first_stage_solution_writer( file_name, scenario, bundling ):
+def first_stage_nonant_npy(file_name, scenario, bundling):
```
bknueven, line 133: I'd be happier with this function name if it had `writer` or `serializer` or something similar in the name

mpisppy/utils/sputils.py
```diff
@@ -510,19 +520,28 @@ def ef_nonants_csv(ef, filename):
     """
     with open(filename, "w") as outfile:
         outfile.write("Node, EF_VarName, Value\n")
-        for (ndname, varname, varval) in ef_nonants(ef):
+        for (ndname, varname, varval) in nonants(ef):
```
bknueven, line 523: This change seems to be responsible for the test failure.

bknueven (commented):
My only real concern is to make our names consistent everywhere if we're replacing "first_stage_solution" with "first_stage_nonant" (or should it be "first_stage_nonants"?).

mpisppy/utils/sputils.py
DLWoodruff: OK. I will replace npy with npy_serializer

bknueven (approved):

PR #117, opened 2021-05-14
MMW

sylvainczx:
Mak, Morton, Wood; Amalgomator; Xhat_Eval


mpisppy/tests/test_ci.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+if '_persistent' in solvername:
+    persistentsolvername = solvername
+else:
+    persistentsolvername = solvername+"_persistent"
+try:
+    persistent_available = pyo.SolverFactory(persistentsolvername).available()
+except:
+    persistent_available = False
```
bknueven, line 46: For maintainability I would prefer we import this logic from test_ef_ph.py or import both from a common source as opposed to copy-pasting it in.

mpisppy/utils/sputils.py
```diff
@@ -508,8 +507,8 @@ def ef_nonants(ef):
     Note:
         not on an EF object because not all ef's are part of an EF object
     """
-    for (ndn,i), var in ef.ref_vars.items():
-        yield (ndn, var, pyo.value(var))
+    for key, val in ef.ref_vars.items():
+        yield (key[0], val, pyo.value(val))
```
bknueven, line 511: I preferred this as-it-was, and it is a non-functioning change.

bknueven (commented):
This all seems very reasonable. I have a few high level comments:
1. In some places when I see "ci" the first thing I think is "continuous integration", not "confidence interval". This is especially true in the testing file names -- they should probably mention "MMW".
2. Following python naming conventions, I think MMWci.py should be renamed mmw_ci.py, and the class `MMWci` should be called `MMW_CI` or `MMWCI` or `MMWConfidenceIntervals` (my preference is the last).
3. xhat_eval is a lot like xhat_tryer. Is there a reason xhat_tryer can't be used?
4. mpisppy/utils will definitely be too crowded if this is merged in. Since MMW is a new high-level feature, I'm not sure it should be hidden away in utils. Obviously not for this PR but the utils directory should probably be re-organized.

mpisppy/tests/test_ci.py
sylvainczx: Done

sylvainczx:
@bknueven High-level comments :
We did 1,2 and 4.
3. (refactor xhat_eval and xhat_tryer) will be done in our next PR.

mpisppy/utils/sputils.py
sylvainczx: Done

PR #119, opened 2021-05-19
Refactor xhat_tryer and xhat_eval

sylvainczx:
Refactor xhat_tryer and xhat_eval, but xhat will not work as a PH extension as written. Something needs to disable W and prox when used as a ph extension.


mpisppy/utils/xhat_eval.py
```diff
@@ -5,21 +5,32 @@
...
 import mpisppy.utils.sputils as sputils
 import mpisppy.spopt
 
     
-fullcomm = mpi.COMM_WORLD
+fullcomm = MPI.COMM_WORLD
 global_rank = fullcomm.Get_rank()
 
 if global_rank==0 :
     print("WHAT ABOUT MULTI-STAGE")
```
bknueven, line 23: Do we need to do some extra work to verify this works with multi-stage fine?

PR #120, opened 2021-05-21
PySP interface test

DLWoodruff:
PySP (plus drop pyutilib from pyomo tracker)

### DIFF
--- a/.github/workflows/testpysp.yml
+++ b/.github/workflows/testpysp.yml
@@ -1,6 +1,6 @@
 # pysp interface (pyomo released)
 
-name: aph tests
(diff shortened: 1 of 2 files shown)


DLWoodruff:
The Schur-complement tests failed for reasons related to github, so I am going to merge.

.github/workflows/testpysp.yml
```diff
@@ -0,0 +1,49 @@
+# pysp interface (pyomo released)
+
+name: aph tests
```
bknueven, line 3: Could we change the name?

PR #122, opened 2021-05-21
Seqsampling

sylvainczx:
Sequential sampling procedures to compute confidence intervals.


doc/src/confidence_intervals.rst
```diff
@@ -1,4 +1,4 @@
-.. _MMW confidence interval:
+.. Confidence intervals:
```
DLWoodruff, line 1: Anchors need a leading underscore

mpisppy/confidence_intervals/seqsampling.py
```diff
@@ -0,0 +1,475 @@
...
+    solvername: str, optional
+        Name of the solver used. The default is "gurobi"
+    solver_options: dict, optional
+        Solving options. The default is None.
+    crops_multiplier: int, optional
+        A parameter of the farmer model. The default is 1.
+
+    Returns
+    -------
+    path: str
```
DLWoodruff, line 49: Probably, this function should return xhat and not write it to a file

mpisppy/utils/xhat_eval.py
```diff
@@ -57,7 +57,8 @@ def __init__(
         #TODO: CHANGE THIS AFTER UPDATE
         self.PH_extensions = None
         
-    
+        self.subproblem_creation(self.verbose)
+        self._create_solvers()
```
bknueven, line 61: The classes derived from `Xhat_Eval` probably also do this somewhere. We should eliminate the associated calls in the derived classes to ensure we're not doing extra work.

bknueven (changes requested):
Need to fix issue with Xhat_Eval spoke friend classes

bknueven (approved):
I fixed the issue I had

PR #131, opened 2021-06-02
adding function for set_instance retries

### LINKED ISSUE #3: Move and generalize license re-try loop logic in PHBase
[#56 in old repository]

JPW: Basic logic should be moved to sputils.py, and the reset logic should be implemented in L-shaped and (maybe) EF, at a minimum.

bknueven:
Fixes #3


DLWoodruff:
Looks good.

PR #133, opened 2021-06-11
multistage for confidence intervals

sylvainczx:
New example mpisppy/tests/examples/aircond_submodels.py
Updating spbase, spopt and various files to run with multistage.
Spbase and subclasses can now support unbalanced trees.


bknueven (approved):
I don't have time for a more careful review, but what's here seems reasonable.

There's perhaps a slight incongruity now within the user interface -- `all_nodenames` is in fact all the nodes, but the list of nodes we ask the user to attach to `_mpisppy_node_list` *does not* have the leaf nodes.

I think this is okay since `_mpisppy_node_list` communicates something about the constraints, but we should probably make sure it's clear an documented.

DLWoodruff (approved):
I made an edit to respond to Ben's comment (more could be done, though), now I am going to merge.

PR #134, opened 2021-06-18
Initial pass at PTDF extension

bknueven:


mpisppy/spopt.py
```diff
@@ -153,7 +153,7 @@ def _vb(msg):
                            np.max(all_set_objective_times)))
 
         if self.extensions is not None:
-            results = self.extobject.pre_solve(s)
+            self.extobject.pre_solve(s)
```
bknueven, line 156:
```suggestion
            results = self.extobject.pre_solve(s)
```

bknueven:
We should test this using the 5bus Prescient example: https://github.com/grid-parity-exchange/Prescient/tree/main/examples/5bus

PR #137, opened 2021-07-13
Multistage sampling & confidence intervals

sylvainczx:


sylvainczx:
- Creating a SampleTree class, and using it to find non-anticipative feasible solution for a whole tree from a first stage policy.
-  Using it to make MMW multistage friendly.
- Reordering the confidence_intervals directory (creation of ciutils.py, refactoring the computation of gap estimators)

DLWoodruff (commented):
This looks very good. I have a handful of things that we can take care of during our next meeting.

PR #140, opened 2021-07-26
A patch for multistage confidence intervals

sylvainczx:
-Preparing sequential sampling for multistage SP
-Gap estimator now works with a single scenario
-Adding documentation to sample_tree.py
-Creating a utility to rescale branching factors


DLWoodruff (approved):
Thanks. Looks good.

PR #142, opened 2021-08-09
Multistage compatibility for xhatshufflelooper spoke

sylvainczx:
1.  Changing the xhatshufflelooper spoke to make it compatible with multistage problems:
- For multistage problems, the spoke shuffles the scenarios, and then loops over the shuffled list (with an increasing starting point) to fill the scenario tree with feasible policies.
- Two attributes ensures a better estimation: reverse (looping over the reversed shuffled list after, to take into account "hidden" scenarios) and iter_step (increasing the starting point between 2 iterations of more than 1, ensuring a better diversity for the first iterations)
2. Adaptation of baseparsers and vanilla to this.
3. Correcting minor typos

### DIFF
--- a/mpisppy/cylinders/xhatshufflelooper_bounder.py
+++ b/mpisppy/cylinders/xhatshufflelooper_bounder.py
@@ -134,7 +134,8 @@ class XhatShuffleInnerBound(spoke.InnerBoundNonantSpoke):
         self.random_stream.seed(self.random_seed)
         #Check that they have the same stream
         check = self.random_stream.normalvariate(0, 1)
-        print(f"Hello, for global rank {global_rank}, random stream gives a value of {check}")
+        chack2 = self.random_stream.normalvariate(0, 1)
+        print(f"Hello, for global rank {global_rank}, random stream gives a value of {check} and {check2}")
         
         # shuffle the scenarios (i.e., sample without replacement)
         shuffled_scenarios = self.random_stream.sample(
--- a/mpisppy/cylinders/xhatshufflelooper_bounder.py
+++ b/mpisppy/cylinders/xhatshufflelooper_bounder.py
@@ -120,6 +120,14 @@ class XhatShuffleInnerBound(spoke.InnerBoundNonantSpoke):
         logger.debug(f"Entering main on xhatshuffle spoke rank {self.global_rank}")
 
         self.xhatbase_prep()
+        if "reverse" in self.opt.options["xhat_looper_options"]:
+            self.reverse = self.opt.options["xhat_looper_options"]["reverse"]
+        else:
+            self.reverse = True
+        if "iter_step" in self.opt.options["xhat_looper_options"]:
+            self.iter_step = self.opt.options["xhat_looper_options"]["iter_step"]
+        else:
+            self.iter_step = None
 
         # give all ranks the same seed
         self.random_stream.seed(self.random_seed)
@@ -128,7 +136,9 @@ class XhatShuffleInnerBound(spoke.InnerBoundNonantSpoke):
         shuffled_scenarios = self.random_stream.sample(self.opt.all_scenario_names, 
                                                        len(self.opt.all_scenario_names))
         scenario_cycler = ScenarioCycler(shuffled_scenarios,
-                                         self.opt.nonleaves)
+                                         self.opt.nonleaves,
+                                         self.reverse,
+                                         self.iter_step)
 
         def _vb(msg): 
             if self.verbose and self.opt.cylinder_rank == 0:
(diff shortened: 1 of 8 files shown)


bknueven (commented):
I left a few minor comments. Overall it seems reasonable.

It would be nice if the `hydro` example was also modified, and perhaps the default settings in baseparsers (e.g., set the default for `with_xhatspecific` to `False`). There's no reason we shouldn't be pointing potential multistage users towards this new capability.

mpisppy/cylinders/xhatshufflelooper_bounder.py
```diff
@@ -168,37 +179,122 @@ def _vb(msg):
...
+        self._cycle_idx = tmp_cycle_idx
+        
+        #Updating scenarios
+        self._cur_ROOTscen = self._shuffled_snames[self._cycle_idx]
+        self.create_nodescen_dict()
+        #TODO Make a better update system for multistage
+        # if self._multi:
+        #     self.nodescen_dict = {'ROOT': self._cur_ROOTscen}
+        # else:
+        #     self.create_nodescen_dict()
```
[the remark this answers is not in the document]
sylvainczx: I modified ScenarioCycler to change this todo into something done, normally decreasing execution time of an item of the loop.

sylvainczx:
I updated, following @bknueven comments. hydro_cylinders is now using xhatshuffle instead of xhatspecific, and with_xhatspecific is set default to False.
Also changed a little bit ScenarioCycler (it was not handling correctly reverse).

examples/hydro/hydro_cylinders.py
```diff
@@ -71,23 +71,18 @@ def main():
...
-                          "ROOT_2": "Scen7"}
     
-    if with_xhatspecific:
-        xhatspecific_spoke = vanilla.xhatspecific_spoke(*beans,
-                                                        xhat_scenario_dict,
+    if with_xhatshuffle:
+        xhatshuffle_spoke = vanilla.xhatshuffle_spoke(*beans,
                                                         all_nodenames,
                                                         BFs,
                                                         scenario_creator_kwargs=scenario_creator_kwargs)
```
bknueven, line 80:
```suggestion
xhatsuffle_spoke["opt_kwargs"]["all_nodenames"] = all_nodenames

```
I'm a bit surprised this works without this. For a separate PR, we should figure out the `beans` issue with `all_nodenames`

mpisppy/utils/vanilla.py
```diff
@@ -287,6 +289,15 @@ def xhatshuffle_spoke(
...
     }
+    if _hasit(args,"add_reversed_shuffle"):
+        xhat_options["xhat_looper_options"]["reverse"] = args.add_reversed_shuffle
+    if _hasit(args,"add_reversed_shuffle"):
+        xhat_options["xhat_looper_options"]["xhatshuffle_iter_step"] = args.xhatshuffle_iter_step
+    
+    if branching_factors:
+        xhat_options["branching_factors"] = branching_factors
+        if all_nodenames is None:
+            all_nodenames = sputils.create_nodenames_from_BFs(branching_factors)
```
sylvainczx, line 300: @bknueven This is related to your last command. If a branching factor is specified without all_nodenames, the spoke creates the all_nodenames

examples/hydro/hydro_cylinders.py
sylvainczx: This is related to the way vanilla treats branching factors and all_nodenames. I made a comment about this in vanilla

mpisppy/utils/vanilla.py
bknueven: Ah, that makes good sense

mpisppy/cylinders/xhatshufflelooper_bounder.py
```diff
@@ -168,37 +179,144 @@ def _vb(msg):
...
+        self._cycle_idx = 0
+        self._cur_ROOTscen = self._shuffled_snames[0]
+        self.create_nodescen_dict()
+        
+        self._scenarios_this_epoch = set()
+    
+    def _begin_reverse_epoch(self):
+        self._reversed = True
+        self._shuffled_snames = list(reversed([s[1] for s in self._shuffled_scenarios]))
+        self._original_order = list(reversed([s[0] for s in self._shuffled_scenarios]))
```
bknueven, line 283:
```suggestion
        self._shuffled_snames = [s[1] for s in reversed(self._shuffled_scenarios)]
        self._original_order = [s[0] for s in reversed(self._shuffled_scenarios)]
```

bknueven (approved):
Having tried this out locally with hydro, I think this is an excellent contribution.

I left a final nit in the comments.

@DLWoodruff this does ask if we can re-introduce the "try best first" logic into the shuffle looper. Looking at the logs for the hydro example, the outer scenarios in the tree generally don't do very well. Alternatively, maybe there could be a "companion spoke" to the shuffle which keeps a list of the "best so far" and tries them. For another PR, obviously.

PR #143, opened 2021-08-11
Amalgamator with cylinders

sylvainczx:
- Update of amalgamator.py to make it work with cylinders and extensions.
- New functions to support the fixer extension in vanilla.
- New writing function for EF that mimic write_spin_the_wheel
- 3 examples farmer_ama.py, uc_ama.py and aircond_ama.py that uses Amalgomator
- Documentation of Amalgomator


bknueven:
It seems like the aircond_ama.py example in runall.py is broken:
```
mpiexec -np 3 python -m mpi4py aircond_ama.py --branching-factors 3 3 --bundles-per-rank=0 --max-iterations=100 --default-rho=1 --with-lagrangian --with-xhatshuffle --solver-name=xpress_persistent
```

mpisppy/utils/amalgomator.py
```diff
@@ -242,29 +341,122 @@ def run(self):
...
+                           "scenario_denouement": self.scenario_denouement,
+                           "all_scenario_names": self.scenario_names,
+                           "scenario_creator_kwargs": self.kwargs}
+            if self.is_multi:
+                    if "all_nodenames" in self.options :
+                        beans["all_nodenames"] = self.options["all_nodenames"]
+                    elif "branching_factors" in self.options:
+                        beans["branching_factors"] = self.options["branching_factors"]
+                    elif "BFs" in self.options:
+                        beans["branching_factors"] = self.options["BFs"]
```
bknueven, line 375: It might be better if this file just creates `all_nodenames` based on the options and passes that down into the hub/spoke creators, as opposed to passing the `branching_factors` to the hub/spoke creators. This may also help address the bug in aircond.

Further (probably not for this PR), I think we could just infer whether the problem is multistage or not by looking for `all_nodenames`, `branching_factors`, and `BFs` in options, and if none are there we assume the model to be 2-stage.

examples/run_all.py
```diff
@@ -183,19 +188,25 @@ def do_one_mmw(dirname, progname, npyfile, efargstring, mmwargstring):
...
        "--solver-name={} --fwph-stop-check-tol 0.01".format(solver_name))
+
 do_one("hydro", "hydro_cylinders.py", 3,
-       "--BFs 3 3 --bundles-per-rank=0 --max-iterations=100 "
+       "--branching-factors 3 3 --bundles-per-rank=0 --max-iterations=100 "
        "--default-rho=1 --with-xhatshuffle --with-lagrangian "
        "--solver-name={}".format(solver_name))
 do_one("hydro", "hydro_cylinders_pysp.py", 3,
-       "--BFs 3 3 --bundles-per-rank=0 --max-iterations=100 "
+       "--branching-factors 3 3 --bundles-per-rank=0 --max-iterations=100 "
```
bknueven, line 197:
```suggestion
       "--bundles-per-rank=0 --max-iterations=100 "
```
`hydro_cylinders_pysp.py` gets its scenario tree from `PySPModel`

PR #147, opened 2021-09-10
Multistage sequential sampling with independent scenarios

sylvainczx:
Creating a file multi_seqsampling.py to extend sequential sampling to multistage.
The class IndepScens_SeqSampling uses gap estimators created using independent scenarios (instead of a scenario tree)


sylvainczx:
Maybe we should:
- Add in the doc that multi_seqsampling must be used for multistage problem
- Or add a warning in seqsampling when dealing with a multistage problem

DLWoodruff (approved):
This has been used extensively already and seems to work well.

PR #148, opened 2021-09-17
making xhat_eval lazily create subproblems and solvers

bknueven:


DLWoodruff (approved):
This is my first experience with something formally called "lazy" creation (although I have done things like this in the past but didn't know it had a name) and I have to say:
- it sure is ugly,
- I think the function name should be changed from _create_subproblem_solvers to _lazy_create_subproblem_solvers, and
- I don't yet understand how this fixes the performance problem, but I think we should merge this because the performance problem it fixes is pretty severe in some cases.

PR #150, opened 2021-09-22
Scenario diagnositics

bknueven:


jeanpaulwatson (commented):
Farmer example should be added to tests.

PR #153, opened 2021-09-30
New models

briancknight:
- adding optional start-up costs for aircond model
- adding GBD model and extended demand data
- reorganizing objective value gap option in mmw_ci
- allowing negative gap realizations (print warning instead of raising RunTimeWarning
- fixing EF_2stage or EF_mstage check in amalgomator


DLWoodruff (approved):
Looks good.

Issue #149, opened 2021-09-22
Questions about getting variable values


yshao44:
Hi there, a quick question, is there a way to get value for variables that not only subject to non-anticipitivity? Please help... Thank you

bknueven:
Yes, but the answer depends on how you're using mpi-sppy.

The `SPBase` objects (`EF`, `PH`, `APH`, `LShaped`) define this method:
https://github.com/Pyomo/mpi-sppy/blob/e1e9d03cc9d883109c41b1175e9a372d6c87ad2c/mpisppy/spbase.py#L623-L630
which will write a tree solution to the directory specified. You can even provide a custom writer for your model. This a appropriate for using these algorithms / EF in a standalone fashion.


If you're using `spin_the_wheel`, then you want to use the below function, which will automatically find the "spoke" with the winning incumbent solution, and calls its `write_tree_solution` method:
https://github.com/Pyomo/mpi-sppy/blob/e1e9d03cc9d883109c41b1175e9a372d6c87ad2c/mpisppy/utils/sputils.py#L186-L194
Again, you can specify a custom writer function for the scenario models.


Lastly, if you're using `mpisppy.sputils.create_EF`, this function will write the tree solution:
https://github.com/Pyomo/mpi-sppy/blob/e1e9d03cc9d883109c41b1175e9a372d6c87ad2c/mpisppy/utils/sputils.py#L596-L607

bknueven:
@DLWoodruff we should add this to the documentation so it is explicit

yshao44:
Thanks for your response. I tested the write_tree_solution() method, and it works for **ef** object. But when solve using benders, it returns the error "no tree solution available". I haven't tested it for **ph**. Could you please verify if write_tree_solution() is correctly implemented for **lshaped** ? Thank you.

bknueven:
I fixed the issue with `LShaped` in the #154; it should have been reporting as having a tree solution available, but it didn't, probably out of an abundance of caution. Note that PH will *only* report it has a solution available if you use one of the included "xhat" PH extensions.

yshao44:
Problem is solved, thank you

PR #154, opened 2021-10-01
Adding tree solution tag for lshaped

bknueven:
#149 points out that LShaped currently does not report a tree solution available at convergence. This PR fixes that.


DLWoodruff (approved):
As advertised.

PR #171, opened 2021-10-28
Add RuntimeError for stale non-anticipative variables on feasible scenarios

### LINKED ISSUE #170: Update_W fails to update dual weights in presence of arbitrary nonanticipative variables
To start, I am running the latest main branch (presently 52cc83152b59ec52d7c8e1e5c1279f56969bcbd8).

In Pyomo it is sometimes convenient to introduce arbitrary decision variables and just let the solver ignore them. I have done this in the attached example -- the variable sales[0] is arbitrary, and I introduced it solely to avoid needing special logic for setting up the root node in the scenario tree.

[inventory.tar.gz](https://github.com/Pyomo/mpi-sppy/files/7428888/inventory.tar.gz)

When I execute `run.sh`, I encounter an issue in the Update_W function of spbase.py:

https://github.com/Pyomo/mpi-sppy/blob/main/mpisppy/phbase.py#L224

The issue is that the arbitrary decision variable is not assigned a value by the solver, so the code attempts to subtract a float from None. It is evident to me that somebody else has already encountered this issue given the commented print lines preceding the subtraction.

Intuitively, if I force sales[0] == 0 (a constraint that is commented out in `inventory.py`), then there is no issue. Unintuitively, the issue also goes away if I change --bundles-per-rank=0 to --bundles-per-rank=1 or --bundles-per-rank=3.

It took me a while to figure out why mpi-sppy was complaining. I suppose it would be nice for such a condition to be automatically detected and handled. Alternatively, it would probably save someone some time down the road if the resulting TypeError were captured and given some additional messaging. Something like "Check that your instance does not have arbitrary decision variables."

bknueven:
Fixes #170. Given that the solve need to touch every variable at least once when loading it into the model, this safety check on *only* the nonanticipative variables once per solve should not be onerous.

### DIFF
--- a/mpisppy/spopt.py
+++ b/mpisppy/spopt.py
@@ -284,13 +284,19 @@ class SPOpt(SPBase):
             if s._mpisppy_data.scenario_feasible:
                 for v in s._mpisppy_data.nonant_indices.values():
                     if v.stale:
-                        raise RuntimeError(
-                                f"Non-anticipative variable {v.name} on scenario {sn} "
-                                 "reported as stale. This usually means this variable "
-                                 "did not appear in any (active) constraints, and hence "
-                                 "was not communicated to the subproblem solver. Please "
-                                 "ensure all non-anticipative variables appear in some "
-                                 "constraint.")
+                        if self.is_zero_prob(s, v) and v._value is None:
+                            raise RuntimeError(
+                                    f"Non-anticipative zero-probability variable {v.name} "
+                                    f"on scenario {sn} reported as stale and has no value. "
+                                     "Zero-probability variables should have a value set.")
+                        else:
+                            raise RuntimeError(
+                                    f"Non-anticipative variable {v.name} on scenario {sn} "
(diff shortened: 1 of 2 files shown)


mpisppy/spopt.py
```diff
@@ -277,6 +277,28 @@ def _vb(msg):
...
+        # we attempt to access the variables `value` attribute (see Issue #170).
+        for sn, s in self.local_scenarios.items():
+            if s._mpisppy_data.scenario_feasible:
+                for v in s._mpisppy_data.nonant_indices.values():
+                    if (not v.fixed) and v.stale:
+                        if self.is_zero_prob(s, v) and v._value is None:
+                            raise RuntimeError(
+                                    f"Non-anticipative zero-probability variable {v.name} "
+                                    f"on scenario {sn} reported as stale and has no value. "
+                                     "Zero-probability variables should have a value set.")
```
bknueven, line 291:
```suggestion
                                     "Zero-probability variables must have a value (e.g., fixed).")
```

PR #179, opened 2021-11-11
Spoke sleep time

### LINKED ISSUE #1: Make SPOKE_SLEEP_TIME automatic
Ben: Probably should be driven by the length of the nonant vector

bknueven:
Closes #1 .

### DIFF
--- a/examples/farmer/farmer_cylinders.py
+++ b/examples/farmer/farmer_cylinders.py
@@ -112,8 +112,6 @@ def main():
     if args.with_xhatshuffle:
         list_of_spoke_dict.append(xhatshuffle_spoke)
 
-    mpisppy.cylinders.SPOKE_SLEEP_TIME = 0.1
-
     wheel = WheelSpinner(hub_dict, list_of_spoke_dict)
     wheel.spin()
 
--- a/examples/hydro/hydro_cylinders.py
+++ b/examples/hydro/hydro_cylinders.py
@@ -14,11 +14,11 @@ import mpisppy.cylinders as cylinders
 
 # For this problem, the subproblems are
 # small and take no time to solve. The
-# default SPOKE_SLEEP_TIME of 0.1 *causes*
+# default SPOKE_SLEEP_TIME of 0.01 *causes*
 # synchronization issues in this case, so
 # we reduce it so as not to dominate the
 # time spent for cylinder synchronization
-cylinders.SPOKE_SLEEP_TIME = 0.0001
+SPOKE_SLEEP_TIME = 0.0001
 
 write_solution = True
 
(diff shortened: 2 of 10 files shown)


dtmildebrath:
I am quite a bit behind so sorry if I'm dredging up old news here. Looks like the sleep_time business is used to make the spokes sleep before they get destroyed at the end of the computation, so that the hub can finalize? Is that right? 

If so, why not just have the hub send a kill signal to the spokes, instead of trying to time it manually?

bknueven:
> I am quite a bit behind so sorry if I'm dredging up old news here. Looks like the sleep_time business is used to make the spokes sleep before they get destroyed at the end of the computation, so that the hub can finalize? Is that right? 
> 
> 
> 
> If so, why not just have the hub send a kill signal to the spokes, instead of trying to time it manually?

Ah, sorry to drag you into this, GitHub suggested you as a reviewer so I clicked it.

At any rate, this thing is used as a heuristic to manage the Spoke ranks becoming out-of-sync with each other. If the hit the buffer too rapidly they'll desynchronize and never come to a clean close, because some will be waiting on some reduction and the others will think it's time to quit.

This PR is an attempt to make the kluge automatic, so the user doesn't need to set anything manually for their problem.

dtmildebrath:
It's as good a reason as any for me to brush back up on what's happening.

So Spoke A is doing whatever, it finishes say 5 min before hub is done. Spoke B finishes 2 min before hub is done. But it breaks because Spoke B died 3 minutes ago?

bknueven:
> So Spoke A is doing whatever, it finishes say 5 min before hub is done. Spoke B finishes 2 min before hub is done. But it breaks because Spoke B died 3 minutes ago?

No -- the issue is with intra-spoke communication. Rank 1 of Spoke A doesn't see the kill signal, but Rank 2 gets there a little bit later and does.

Now that Spoke is deadlocked because Rank 2 is waiting to quit and Rank 1 is waiting for a reduction or something else from Rank 2.

By slowing down how often the spoke ranks check the buffer and see if it's time to quit, we (mostly) a avoid these deadlocks.

dtmildebrath:
Okay that makes a lot more sense. Maybe intentionally naive question: why doesn't spoke A.1 see the kill signal?

dtmildebrath:
I'm assuming it's because A.1 checked the hub's "kill it" buffer before the hub wrote the kill signal to it?

bknueven:
> I'm assuming it's because A.1 checked the hub's "kill it" buffer before the hub wrote the kill signal to it?

Yeah, or there was some communication delay. We seemed to be able to consistently reproduce this when the Spokes had trivial optimization problems, and were just constantly checking the buffer.

In general, we think you could run into this issue if the Spoke's optimization problems are sufficiently different in execution time and/or are much faster than the Hub. Hence the desire to try to tie the delay to problem size.

bknueven:
@DLWoodruff We probably need to change the error message Lagrangian prints out when it notices its ranks are out-of-sync to direct the user to the new option (and maybe print the current value).

dtmildebrath:
But this should only happen if A.1 needs something from A.2 right?

bknueven:
> But this should only happen if A.1 needs something from A.2 right? 

Sure, but this is always the case. All of the spokes minimally use a sum reduce to compute the bound they're working on, I think.

DLWoodruff (approved):

PR #180, opened 2021-11-14
Doc for Pr179

DLWoodruff:


mpisppy/cylinders/lagrangian_bounder.py
```diff
@@ -49,7 +49,8 @@ def lagrangian(self):
         if total == serial_number_sum:
             return bound
         elif self.cylinder_rank == 0:
-            print("WARNING: Lagrangian spokes out of snyc, consider changing mpisppy.cylinders.SPOKE_SLEEP_TIME")
+            print("WARNING: Lagrangian spokes out of snyc, consider changing the spoke_sleep_time option; "
+                  "see the documentation for more information")
```
bknueven, line 53: We should print the current value here, so the user doesn't have to guess at what it is.

bknueven (approved):
I can open an issue for my comment, if you want to merge this as-is.

mpisppy/cylinders/lagrangian_bounder.py
DLWoodruff: good point

PR #190, opened 2021-12-15
change python version on pyomo tracker test

DLWoodruff:


bknueven:
Python 3.9 did not work because the ["anaconda" version of mpi4py](https://anaconda.org/anaconda/mpi4py) is now over a year old. We should probably switch over to the ["conda-forge" version](https://anaconda.org/conda-forge/mpi4py), which supports more platforms, including Windows and ARM-based Macs.

PR #204, opened 2022-01-25
repair langranger rho update factors

DLWoodruff:


bknueven (approved):
I don't see any issues here.

Question: could the Lagranger use the [norm rho updater](https://github.com/Pyomo/mpi-sppy/blob/main/mpisppy/extensions/norm_rho_updater.py)? Could make a lot of sense with some custom options for dual convergence.

PR #205, opened 2022-01-27
Trying multistage FWPH

bknueven:
This runs. Questions:
1. Is the bound provided valid?
1. Is our implementation accidentally *the* multistage implementation?


DLWoodruff:
Line 35 of straight_tests.py needs to have --no-fwph added to it (the alternative is to change -np 3 to -np 4 in the same line)

dtmildebrath:
I don't suppose we have any other 3-stage IPs lying around? (I vaguely recall a forestry problem...?)

dtmildebrath:
Skimmed back through the Boland paper. I would have to set pen to paper to be confident in this, but it seems more plausible to me now that this may be _the_ multistage algorithm. However, as Ben as already raised, it's still possible that this is a weaker version that it might be. In any event, I'm about 90% convinced this should produce valid lower bounds for multistage problems.

bknueven:
Doing my own review, I'm inclined to agree with @dtmildebrath. The only complicating part would be in computing xbar (or z in the Boland paper), but this is already handled by `PHBase`. The FWPH implementation we have separates the variables by non-leaf variables and leaf variables, which is exactly what it seems you would want to do for a multistage extension of this algorithm.

Therefore I've removed the warning when using FWPH with multistage, and I think this should be merged.

DLWoodruff (approved):

PR #224, opened 2022-02-17
add a little bit of doc for slicing when adding nonants

DLWoodruff:


bknueven (approved):
I changed this to use a list -- we can pass singletons in, but I think that's an undocumented feature. No reason to make this more complicated in explanation.

PR #246, opened 2022-04-12
Global config

DLWoodruff:
A non-disruptive proposal for switching to the pyomo config system.
utils.config.py has the config object
utils.cfg_vanilla.py is a version of vanilla.py that is compatible.
examples.hydro.hydro_cylinders_config.py is an example that uses them and examples.hydro.cfg.bash runs that example

### DIFF
--- a/mpisppy/utils/config.py
+++ b/mpisppy/utils/config.py
@@ -1,0 +1,525 @@
+# This software is distributed under the 3-clause BSD License.
+# Started 12 April by DLW
+# Replace baseparsers.py and enhance functionality.
+# The vision is that global_config ConfigDict can be accessed easily
+# by all code.
+# NOTE: at first I am just copying directly from baseparsers.py
+# TBD: reorganize some of this.
+
+""" Notes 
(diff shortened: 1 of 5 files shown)


mpisppy/utils/config.py
```diff
@@ -0,0 +1,536 @@
...
+import argparse
+import pyomo.common.config as pyofig
+
+global_config = pyofig.ConfigDict()
+
+def _common_args():
+    raise RuntimeError("_common_args is no longer used. See comments at top of config.py")
+
+def popular_args():
+    global_config.declare("max-iterations", pyofig.ConfigValue(
```
bknueven, line 66:
```suggestion
    global_config.declare("max_iterations", pyofig.ConfigValue(
```

This is better-looking python (e.g., `global_config["max_iterations"]`), and Pyomo config system is smart enough to convert it to `--max-iterations` for the command line.

jeanpaulwatson (approved):
Only adds new code - seems innocuous.

PR #260, opened 2022-07-27
Storing and restoring tree solution for incumbent finders

### LINKED ISSUE #256: Incumbent finders should store their full tree solution with each new incumbent
This trouble is sporadic, but sometimes can be seen with netdes 10-30-L-01 with cross-scenario)

bknueven:
Fixes #256


DLWoodruff (approved):
This is a good change and I guess my overly pedantic request for an option to have the old behavior should wait until we have a use-case that needs it, which seems unlikely. BTW: I am going to approve this with the unrelated test failure (an "optional" package that causes Pyomo to fail in an unrelated test) then do a PR to fix the test.

PR #262, opened 2022-08-12
add an option to disable the popular-with-some toc Initializing SPBase

DLWoodruff:
This is harmless, but helpful to me

### DIFF
--- a/mpisppy/spbase.py
+++ b/mpisppy/spbase.py
@@ -86,7 +86,8 @@ class SPBase:
         self.n_proc = self.mpicomm.Get_size()
         self.global_rank = MPI.COMM_WORLD.Get_rank()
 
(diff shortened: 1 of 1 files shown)


mpisppy/spbase.py
```diff
@@ -86,7 +86,8 @@ def __init__(
         self.n_proc = self.mpicomm.Get_size()
         self.global_rank = MPI.COMM_WORLD.Get_rank()
 
-        global_toc("Initializing SPBase")
+        if not options.get("no_toc", False):
```
bknueven, line 89:
```suggestion
        if options.get("toc", True):
```
I don't really like negative options ...

DLWoodruff: I do... but I'll let you win this one! (You are almost surely in the majority)

PR #264, opened 2022-08-24
Rename slamup/slamdown slammax/slammin

bknueven:
@jeanpaulwatson and I agree that these names are slightly more appropriate, as we don't slam to a bound.


DLWoodruff (approved):

DLWoodruff:
No docs?

PR #279, opened 2022-10-04
mmw comm

xiaotie-chen:
mmw can be passed an mpi comm


DLWoodruff:
The schur-complement test is failing for reasons related to test infrastructure, so I am going to merge this.

PR #280, opened 2022-10-25
fix bugs in the writer for ef first stage Vars

DLWoodruff:


bknueven (approved):
Copy+paste is always a bit dangerous ...

PR #305, opened 2023-04-06
swtiching schur yml to ubuntu latest

DLWoodruff:


bknueven (approved):
Ah, that was the problem with this one!

PR #313, opened 2023-06-01
Pyotracker fix

DLWoodruff:


.github/workflows/pyotracker.yml
```diff
@@ -17,19 +17,20 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        pyomo-version: [6.4.2, main]
+        pyomo-version: [6.6.1, main]
```
bknueven, line 20: Are we no longer compatible with Pyomo 6.4.2?

The original idea here was to try to cover the oldest release we are compatible with to the bleeding edge.

DLWoodruff: OK. I'll switch it back to 6.4.2

DLWoodruff: Ben,  Can you update the doc test to Python 3.9?

PR #324, opened 2023-08-02
Uniform ef

DLWoodruff:
make it so EF formation support "uniform" probability specification (and update a sample bash script)

### DIFF
--- a/mpisppy/utils/sputils.py
+++ b/mpisppy/utils/sputils.py
@@ -193,7 +193,7 @@ def create_EF(scenario_names, scenario_creator, scenario_creator_kwargs=None,
     probs_specified = \
         all([hasattr(scen, '_mpisppy_probability') for scen in scen_dict.values()])
     uniform_specified = \
-        all([scen._mpisppy_probability == "uniform" for scen in scen_dict.values()])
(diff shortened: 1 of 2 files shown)


mpisppy/utils/sputils.py
```diff
@@ -192,10 +192,12 @@ def create_EF(scenario_names, scenario_creator, scenario_creator_kwargs=None,
     # Check if every scenario has a specified probability
     probs_specified = \
         all([hasattr(scen, '_mpisppy_probability') for scen in scen_dict.values()])
-    if not probs_specified:
+    uniform_specified = \
+        probs_specified and all([scen._mpisppy_probability == "uniform" for scen in scen_dict.values()])
```
bknueven, line 196: - Since we changed this line -- we don't need to construct the intermediate list for `all`:
```suggestion
        probs_specified and all((scen._mpisppy_probability == "uniform") for scen in scen_dict.values())
```

Similar logic applied to line 194.

bknueven (approved):

mpisppy/utils/sputils.py
DLWoodruff: good catch, good point, and thanks!

PR #326, opened 2023-08-11
gradient rho improvements

ulysse-n:


mpisppy/convergers/norms_and_residuals.py
```diff
@@ -0,0 +1,156 @@
...
+    global_sum = np.zeros(1)
+    for s in PHB.local_scenarios.values():
+        for node in s._mpisppy_node_list:
+            ndn = node.name
+            nlen = s._mpisppy_data.nlens[ndn]
+            xbars = np.fromiter((s._mpisppy_model.xbars[ndn,i]._value
+                                  for i in range(nlen)), dtype='d')
+            nonants_array = np.fromiter((v._value for v in node.nonant_vardata_list),
+                                        dtype='d', count=nlen)
+            scaled_norm = np.divide(np.abs(nonants_array - xbars), xbars)
```
bknueven, line 53: We need to guard against one of the terms of `xbars` being zero.

mpisppy/convergers/norms_and_residuals.py
```diff
@@ -0,0 +1,156 @@
...
+            if not s._mpisppy_data.has_variable_probability:
+                local_sum[0] += s._mpisppy_data.prob_coeff[ndn] * np.sum(np.multiply(scaled_norm, scaled_norm))
+            else:
+                # rarely-used overwrite in the event of variable probability
+                prob_array = np.fromiter((s._mpisppy_data.prob_coeff[ndn_i[0]][ndn_i[1]]
+                                          for ndn_i in s._mpisppy_data.nonant_indices
+                                          if ndn_i[0] == ndn),
+                                         dtype='d', count=nlen)
+                local_sum[0] += np.dot(prob_array, scaled_norm)
```
bknueven, line 62: This term doesn't have the squared scaled norm

mpisppy/convergers/norms_and_residuals.py
```diff
@@ -0,0 +1,156 @@
...
+def scaled_dual_metric(PHB, w_cache, curr_iter):
+    """
+    Compute the Compute the scaled norm of the difference between to consecutive Ws.
+    """
+    local_sum = np.zeros(1)
+    global_sum = np.zeros(1)
+    for sname, s in PHB.local_scenarios.items():
+        current_w = np.array(w_cache[curr_iter][sname])
+        prev_w = np.array(w_cache[curr_iter-1][sname])
+        scaled_norm = np.divide(np.abs(current_w - prev_w), np.abs(current_w))
```
bknueven, line 78: Could be dividing by zero...

mpisppy/convergers/norms_and_residuals.py
```diff
@@ -0,0 +1,156 @@
...
+            nlen = s._mpisppy_data.nlens[ndn]
+            if not s._mpisppy_data.has_variable_probability:
+                local_sum[0] += s._mpisppy_data.prob_coeff[ndn] * np.sum(np.multiply(scaled_norm, scaled_norm))
+            else:
+                # rarely-used overwrite in the event of variable probability
+                prob_array = np.fromiter((s._mpisppy_data.prob_coeff[ndn_i[0]][ndn_i[1]]
+                                          for ndn_i in s._mpisppy_data.nonant_indices
+                                          if ndn_i[0] == ndn),
+                                         dtype='d', count=nlen)
+                local_sum[0] += np.dot(prob_array, scaled_norm)
```
bknueven, line 90: Scaled norm isn't squared.

mpisppy/convergers/norms_and_residuals.py
```diff
@@ -0,0 +1,156 @@
...
+                local_sum[0] += s._mpisppy_data.prob_coeff[ndn] * np.sum(np.multiply(resid, resid))
+            else:
+                # rarely-used overwrite in the event of variable probability
+                prob_array = np.fromiter((s._mpisppy_data.prob_coeff[ndn_i[0]][ndn_i[1]]
+                                          for ndn_i in s._mpisppy_data.nonant_indices
+                                          if ndn_i[0] == ndn),
+                                         dtype='d', count=nlen)
+                local_sum[0] += np.dot(prob_array, resid)
+    PHB.comms["ROOT"].Allreduce(local_sum, global_sum, op=MPI.SUM)
+    return np.sqrt(global_sum[0])
```
bknueven, line 152: These are largely similar to the functions above. I think they should be combined with a flag for scaling or not.

mpisppy/extensions/gradient_extension.py
```diff
@@ -40,60 +41,80 @@ def __init__(self, opt, comm=None):
...
+        self.grad_object.write_grad_rho()
+        rho_setter_kwargs = self.opt.options['rho_setter_kwargs'] \
+                            if 'rho_setter_kwargs' in self.opt.options \
+                            else dict()
+        for sname, scenario in self.opt.local_scenarios.items():
+            rholist = self.rho_setter(scenario, **rho_setter_kwargs)
+            for (vid, rho) in rholist:
+                (ndn, i) = scenario._mpisppy_data.varid_to_nonant_index[vid]
+                scenario._mpisppy_model.rho[(ndn, i)] = rho
+        self.display_rho_values()
```
bknueven, line 60: Is there an option to turn display off, other than hacking the code?

Maybe we should be using loggers...

mpisppy/extensions/gradient_extension.py
```diff
@@ -40,60 +41,80 @@ def __init__(self, opt, comm=None):
...
+
+
+    def _rho_primal_crit(self):
+        primal_thresh = 0.001 # or other value
```
bknueven, line 64: option drive...

mpisppy/extensions/gradient_extension.py
```diff
@@ -40,60 +41,80 @@ def __init__(self, opt, comm=None):
...
+        self.prev_primal_norm = self.curr_primal_norm
+        self.curr_primal_norm = norms.scaled_primal_metric(self.opt)
+        norm_diff = np.abs(self.curr_primal_norm - self.prev_primal_norm)
+        #print(f'{norm_diff =}')
+        return (norm_diff <= primal_thresh)
+
+    def _rho_dual_crit(self):
+        dual_thresh = 0.1 # or other value
```
bknueven, line 72: option drive..

mpisppy/extensions/gradient_extension.py
```diff
@@ -40,60 +41,80 @@ def __init__(self, opt, comm=None):
...
+        self.wt.grab_local_Ws()
+        dual_norm = norms.scaled_dual_metric(self.opt, self.wt.local_Ws, self.opt._PHIter)
+        #print(f'{dual_norm =}')
+        return (dual_norm <= dual_thresh)
+
+    def _rho_primal_dual_crit(self):
+        pd_thresh = 2 #or other value
```
bknueven, line 79: option drive ...

mpisppy/extensions/gradient_extension.py
```diff
@@ -40,60 +41,80 @@ def __init__(self, opt, comm=None):
...
+        self.prev_primal_norm = self.curr_primal_norm
+        self.curr_primal_norm = norms.scaled_primal_metric(self.opt)
+        norm_diff = np.abs(self.curr_primal_norm - self.prev_primal_norm)
+        #print(f'{norm_diff =}')
+        return (norm_diff <= primal_thresh)
+
+    def _rho_dual_crit(self):
+        dual_thresh = 0.1 # or other value
+        self.wt.grab_local_Ws()
+        dual_norm = norms.scaled_dual_metric(self.opt, self.wt.local_Ws, self.opt._PHIter)
```
bknueven, line 74: `_rho_primal_crit` is tracking changes in primal norm, but `_rho_dual_crit` is just tracking the absolute dual norm

mpisppy/extensions/gradient_extension.py
```diff
@@ -40,60 +41,80 @@ def __init__(self, opt, comm=None):
...
+        #print(f'{dual_norm =}')
+        return (dual_norm <= dual_thresh)
+
+    def _rho_primal_dual_crit(self):
+        pd_thresh = 2 #or other value
+        self.wt.grab_local_xbars()
+        primal_resid = norms.primal_residuals_norm(self.opt)
+        dual_resid = norms.dual_residuals_norm(self.opt, self.wt.local_xbars, self.opt._PHIter)
+        resid_rel_norm = dual_resid / primal_resid
```
bknueven, line 83: zero division

bknueven (commented):
We really should minimally address zero division errors before merging

mpisppy/convergers/norms_and_residuals.py
DLWoodruff: These functions could/should be factored, but it not quite as simple as a flag for scaling, so I am not going to do it right now.

mpisppy/extensions/gradient_extension.py
DLWoodruff: I'm adding options, but this code does not seem to be tested.

mpisppy/extensions/gradient_extension.py
DLWoodruff: I updated the help for the config items for the thresholds.

examples/uc/gradient_uc_cylinders.py
```diff
@@ -0,0 +1,240 @@
...
+    if cross_scenario_cuts:
+        ext_classes.append(CrossScenarioExtension)
+    if cfg.xhat_closest_tree:
+        ext_classes.append(XhatClosest)
+    
+    if track_ws:
+        results_dir_name="gradient_uc_cylinders_tracking_W"
+        create_dir_structure(results_dir_name, hub_dict)
+
+    print("dropping ph_ob!!!")
```
gcstinchfield, line 124: Is there a reason this is printed explicitly? Might be better just as a comment

examples/uc/gradient_uc_cylinders.py
```diff
@@ -0,0 +1,240 @@
...
+        list_of_spoke_dict.append(fw_spoke)
+    if lagrangian:
+        list_of_spoke_dict.append(lagrangian_spoke)
+    if xhatlooper:
+        list_of_spoke_dict.append(xhatlooper_spoke)
+    if xhatshuffle:
+        list_of_spoke_dict.append(xhatshuffle_spoke)
+    if cross_scenario_cuts:
+        list_of_spoke_dict.append(cross_scenario_cuts_spoke)
+    print("dropping ph_ob!!!")
```
gcstinchfield, line 221: Same question as before, why print this here?

mpisppy/convergers/norms_and_residuals.py
```diff
@@ -0,0 +1,159 @@
...
+            if not s._mpisppy_data.has_variable_probability:
+                local_sum[0] += s._mpisppy_data.prob_coeff[ndn] * np.sum(np.multiply(scaled_norm, scaled_norm))
+            else:
+                # rarely-used overwrite in the event of variable probability
+                prob_array = np.fromiter((s._mpisppy_data.prob_coeff[ndn_i[0]][ndn_i[1]]
+                                          for ndn_i in s._mpisppy_data.nonant_indices
+                                          if ndn_i[0] == ndn),
+                                         dtype='d', count=nlen)
+                ### TBD: check this!!
+                local_sum[0] += np.dot(prob_array, scaled_norm^2)
```
gcstinchfield, line 63: Will this work? In all of my experience in python, ^ never works for a power, you always need **
I ask because the comment implies this may never have been tested.

DLWoodruff: This code is in a block for variable probability, which is not tested much and so far, not used at all with the converger. I changed it to **2

gcstinchfield (approved):

mpisppy/utils/config.py
```diff
@@ -625,33 +625,46 @@ def gradient_args(self):
                            description="name of the gradient rho file (must be csv)",
                            domain=str,
                            default='')
-        self.add_to_config("order_stat",
-                           description="order statistic for rho (must be between 0 and 1)",
+        self.add_to_config("grad_display_rho",
+                           description="display rho during gradient calscs (default True)",
```
jeanpaulwatson, line 629: Typo in doc string needs to be fixed - "calscs" instead of "calcs".

mpisppy/utils/config.py
```diff
@@ -625,33 +625,46 @@ def gradient_args(self):
...
+                           domain=float,
+                           default=0.1)
+        self.add_to_config("grad_pd_thresh",
+                           description="threshold for dual/primal during gradient calcs",
+                           domain=float,
+                           default=0.1)
+        self.add_to_config("grad_order_stat",
+                           description="order statistic for gradient based rho (must be between 0 and 1)",
                            domain=float,
                            default=-1.0)
```
jeanpaulwatson, line 647: Shouldn't the default value for the order statistic be valid? (as in on the interval (0,1))?

mpisppy/utils/config.py
```diff
@@ -625,33 +625,46 @@ def gradient_args(self):
...
-        self.add_to_config("rho_path",
-                           description="csv file for the rho setter",
+        
+        self.add_to_config("grad_rho_path",
+                           description="csv file for the the grad based rho file (???)",
                            domain=str,
                            default='')
-        self.add_to_config("order_stat",
+        self.add_to_config("grad_order_stat",
                            description="order statistic for rho: must be between 0 (the min) and 1 (the max); 0.5 iis the average",
```
jeanpaulwatson, line 664: "iis" should be changed to "is".

mpisppy/utils/config.py
```diff
@@ -625,33 +625,46 @@ def gradient_args(self):
...
-        self.add_to_config('rho_setter',
+        self.add_to_config('grad_rho_setter',
                            description="use rho setter from a rho file",
                            domain=bool,
                            default=False)
-        self.add_to_config("rho_path",
-                           description="csv file for the rho setter",
+        
+        self.add_to_config("grad_rho_path",
+                           description="csv file for the the grad based rho file (???)",
```
jeanpaulwatson, line 660: Why the "???"?  - Also, fix "the the". Also, it's a directory - no (as in no file)?

mpisppy/utils/gradient.py
```diff
@@ -104,8 +113,8 @@ def find_grad_cost(self):
                     for v in node.nonant_vardata_list:
                         v.unfix()
 
-            grad_cost ={sname: self.compute_grad(sname, scenario) 
-                        for sname, scenario in self.ph_object.local_scenarios.items()} 
+            grad_cost = {sname: self.compute_grad(sname, scenario)
```
jeanpaulwatson, line 116: Shouldn't the "grad_cost" be "grad_dict"? I believe grad_cost is gone...

mpisppy/phbase.py
```diff
@@ -810,17 +811,10 @@ def _vb(msg):
 
         self._update_E1()  # Apologies for doing this after the solves...
         if (abs(1 - self.E1) > self.E1_tolerance):
-            if self.cylinder_rank == 0:
-                print("ERROR")
-                print("Total probability of scenarios was ", self.E1)
-                print("E1_tolerance = ", self.E1_tolerance)
-            quit()
+            raise RuntimeError(f"Total probability of scenarios was {self.E1};  E1_tolerance = ", self.E1_tolerance)
```
jeanpaulwatson, line 814: Don't you want to keep the cylinder rank 0 check? To avoid every rank throwing the exception? The runtime error is definitely better than the quit().

mpisppy/phbase.py
```diff
@@ -810,17 +811,10 @@ def _vb(msg):
...
         feasP = self.feas_prob()
         if feasP != self.E1:
-            if self.cylinder_rank == 0:
-                print("ERROR")
-                print("Infeasibility detected; E_feas, E1=", feasP, self.E1)
-            quit()
+            raise RuntimeError(f"Infeasibility detected; E_feas={feasp}, E1={self.E1}")
```
jeanpaulwatson, line 817: Same comment as above here (why not retain the cylinder rank 0 check).

examples/farmer/farmer_rho_demo.bash
```diff
@@ -2,7 +2,11 @@
 
 SOLVERNAME="cplex"
 
-mpiexec -np 3 python -m mpi4py farmer_rho_demo.py --num-scens 3 --bundles-per-rank=0 --max-iterations=1000 --default-rho=1 --solver-name=${SOLVERNAME} --xhatpath=./_farmer_cyl_nonants.npy --order-stat 0.5 --xhatshuffle --lagrangian --max-stalled-iters 5000 --rho-setter --rel-gap 0.001
+# get an xhat
+# xhat output file name is hardwired to 'farmer_cyl_nonants.npy'
+mpiexec -np 3 python -m mpi4py farmer_cylinders.py  --num-scens 3 --lagrangian --xhatshuffle --bundles-per-rank=0 --max-iterations=50 --default-rho=1 --solver-name=${SOLVERNAME}
+
+mpiexec -np 3 python -m mpi4py farmer_rho_demo.py --num-scens 3 --bundles-per-rank=0 --max-iterations=100 --default-rho=1 --solver-name=${SOLVERNAME} --xhatpath=./farmer_cyl_nonants.npy --grad-order-stat 0.5 --xhatshuffle --lagrangian --max-stalled-iters 5000 --grad-rho-setter --rel-gap 0.001
```
jeanpaulwatson, line 9: Nit - but for consistency - shouldn't --xhatpath really be --xhatfile?

examples/farmer/farmer_rho_demo.py
```diff
@@ -72,14 +76,16 @@ def main():
...
+    rho_setter = None  # non-grad rho setter?
 
     if cfg.default_rho is None and rho_setter is None:
         raise RuntimeError("No rho_setter so a default must be specified via --default-rho")
 
     if cfg.use_norm_rho_converger:
         if not cfg.use_norm_rho_updater:
             raise RuntimeError("--use-norm-rho-converger requires --use-norm-rho-updater")
+        elif cfg.grad_rho_setter:
+            raise RuntimeError("You cannot have--use-norm-rho-converger and --grad-rho-setter")
```
jeanpaulwatson, line 88: Largely unproductive whining, which I should put in an issue - we should be consistent and have "--norm-rho-converger" without the "--use" prefix.

examples/farmer/farmer_rho_demo.py
```diff
@@ -72,14 +76,16 @@ def main():
 
     num_scen = cfg.num_scens
     crops_multiplier = cfg.crops_mult
-    rho_setter = None
+    rho_setter = None  # non-grad rho setter?
```
jeanpaulwatson, line 79: I think you can drop the "?" - I believe you are correct.

mpisppy/cylinders/xhatxbar_bounder.py
```diff
@@ -58,11 +58,7 @@ def ib_prep(self):
 
         self.opt._update_E1()  
         if (abs(1 - self.opt.E1) > self.opt.E1_tolerance):
-            if self.opt.cylinder_rank == 0:
-                print("ERROR")
-                print("Total probability of scenarios was ", self.opt.E1)
-                print("E1_tolerance = ", self.opt.E1_tolerance)
-            quit()
+            raise RuntimeError(f"Total probability of scenarios was {self.E1};  E1_tolerance = ", self.E1_tolerance)
```
jeanpaulwatson, line 61: Same comment I raised elsewhere - why can is the cylinder_rank=0 check dropped?

jeanpaulwatson (approved):
Minor comments - likely only a few things that are must-fix (I think typos).

mpisppy/utils/config.py
DLWoodruff: He wants to require value be given, so he flagged it. The default should be None, really (but then the test for a missing entry where the value is used needs to be changed). The alternative is to set the default to 0.5 and drop the test for a missing entry. If I get time, I will do the former; but I won't get time until tomorrow at the earliest.

mpisppy/utils/config.py
DLWoodruff: The ??? is to remind me that I need to find out what actually goes in this file and document it in this help string

mpisppy/phbase.py
DLWoodruff: It is not clear to me whether the rank 0 check (which was my idea in this case) was a good idea. I think you want every rank to shut down if infeasibility is detected rather than relying on the user to have done -m mpy4py so it will shut everyone down

examples/farmer/farmer_rho_demo.bash
DLWoodruff: He has both and I have been meaning to sort this out.

examples/farmer/farmer_rho_demo.py
DLWoodruff: I agree

examples/farmer/farmer_rho_demo.py
DLWoodruff: I agree

mpisppy/cylinders/xhatxbar_bounder.py
DLWoodruff: Same reply: It is not clear to me whether the rank 0 check (which was my idea in this case) was a good idea. I think you want every rank to shut down if infeasibility is detected rather than relying on the user to have done -m mpy4py so it will shut everyone down

mpisppy/utils/config.py
DLWoodruff: done

PR #329, opened 2023-08-30
Add Primal/Dual Convergence metrics

garg02:


mpisppy/convergers/primal_dual_converger.py
```diff
@@ -0,0 +1,118 @@
...
+        Get the current xbar values from the local scenarios
+        Returns:
+            xbars (dict): dictionary of xbar values indexed by 
+                          (decision node name, index)
+        """
+        xbars = {}
+        for s in self._ph.local_scenarios.values():
+            for ndn_i, xbar in s._mpisppy_model.xbars.items():
+                xbars[ndn_i] = xbar.value
+        return xbars
```
bknueven, line 39:
```suggestion
        for s in self._ph.local_scenarios.values():
            for ndn_i, xbar in s._mpisppy_model.xbars.items():
                xbars[ndn_i] = xbar.value
            break
        return xbars
```

mpisppy/convergers/primal_dual_converger.py
```diff
@@ -0,0 +1,118 @@
...
+import numpy as np
+import mpisppy.convergers.converger
+from mpisppy import MPI
+
+class PrimalDualConverger(mpisppy.convergers.converger.Converger):
+    """ Convergence checker for the primal-dual metrics.
+        Primal convergence is measured as weighted sum over all scenarios s
+        p_{s} * ||x_{s} - \bar{x}||_1.
+        Dual convergence is measured as 
+        rho * ||\bar{x}_{t} - \bar{x}_{t}||_1 * r_{s}
```
bknueven, line 11:
```suggestion
        rho * ||\bar{x}_{t} - \bar{x}_{t-1}||_1
```

mpisppy/convergers/primal_dual_converger.py
```diff
@@ -0,0 +1,118 @@
...
+            self (object): create by prep
+
+        Returns:
+           converged?: True if converged, False otherwise
+        """
+
+        primal_gap = self._compute_primal_convergence()
+        dual_gap = self._compute_dual_residual()
+        self.prev_xbars = self._get_xbars()
+        ret_val = primal_gap + dual_gap <= self.convergence_threshold
```
bknueven, line 107: Why this? And not:

```suggestion
        ret_val = max(primal_gap, dual_gap) <= self.convergence_threshold
```
?

bknueven (commented):
A few minor comments

mpisppy/convergers/primal_dual_converger.py
garg02: I don't have a good rationale, except that is what the original PySP implementation had. In practice, one of the terms is 10-100x the other one, so it is effectively a max and we can do that

DLWoodruff: Ben makes the case that max is easier to reason about when setting the convergence threshold, so please make the change.

DLWoodruff (approved):

PR #330, opened 2023-08-30
[WIP] Agnostic

DLWoodruff:
Notes on Sept 4, 2023; updated Sept 7
- PH and xhatshuffle seem to be working. Lagrangian executes, but does not seem to work correctly (I have not had time to look into it)
- I have some tests
- After fixing lagrangian the next thing would be look at an EF "template" or some form of programmable support for the EF so bundles can be used
- There are examples (but definitely not examplars) for pyomo and AMPL in the farmer directory with bash scripts to drive them.
- AMPL PH seems to work correctly, except that the obj value that it keeps internally (for historic reasons) does not match the value that is kept when we use pyomo. This doesn't really matter (the upper bound from the xhatter is correct), but it bothers me so I am leaving the test as-is so if fails.


mpisppy/phbase.py
```diff
@@ -269,6 +271,8 @@ def __init__(
         # Note that options can be manipulated from outside on-the-fly.
         # self.options (from super) will archive the original options.
         self.options = options
+        self.Ag = options.get("Ag", None)  # The Agnostic Object
```
bknueven, line 274:
```suggestion
        class _MockAgnostic:
            def callout_agnostic(self, *args, **kwargs):
                pass
 
        self.Ag = options.get("Ag", _MockAgnostic)  # The Agnostic Object
```

PR #333, opened 2023-09-06
pass hub bounds to spoke for all spoke types

garg02:


bknueven (commented):
I'll need to do a more careful review later.

My meta-comment is that I think we should eliminate the option-driving here and just always send the inner and outer bounds to the spokes.

We already send at least one double for the poison pill / serial number, sending two more won't make a meaningful performance difference. And it will significantly simplify the code.

bknueven (approved):
Thanks! I like the simplification

PR #334, opened 2023-09-08
PHTracking extension

garg02:
This PR is 98% independent of #333, but until sending hub bounds to spoke is finalized, convergence tracking that require hub bounds won't work. I have disabled the specific tracking args in cfg_vanilla.py so the tests will run fine.

Notes of changes that have been made by file
phtracker.py:
- Implemented tracking as an extension;
- As some of the tracking code (bounds and nonants) could be used for other optimizers besides PH,  maybe we should make subclasses of TrackedData for each type. Thoughts?

cfg_vanilla:
- Used extension_adder to simplify the process of bridging command line arguments with ph options

farmer_cylinders.py: 
- using extension_adder so extensions can also be added from cfg_vanilla

primal_dual_converger.py
- Implemented tracking

lagranger_bounder.py and lagrangian_bounder.py
- initialize iteration variable earlier so tracker can access
- call extensions post_loops()

converger.py, extension.py, phbase.py, and spopt.py
- added hooks to call extension and converger methods

config.py
- added command line arguments to use wxbar reader and writer and PHTracker

examples/run_all.py
- testing changes

### DIFF
--- a/mpisppy/cylinders/lagranger_bounder.py
+++ b/mpisppy/cylinders/lagranger_bounder.py
@@ -93,13 +93,13 @@ class LagrangerOuterBound(mpisppy.cylinders.spoke.OuterBoundNonantSpoke):
 
         self.lagrangian_prep()
 
+        self.A_iter = 1
         self.trivial_bound = self._lagrangian(0)
 
         self.bound = self.trivial_bound
 
         self.opt.current_solver_options = self.opt.iterk_solver_options
 
-        self.A_iter = 1
         while not self.got_kill_signal():
             # because of aph, do not check for new data, just go for it
             self.bound = self._update_weights_and_solve(self.A_iter)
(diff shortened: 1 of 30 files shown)


mpisppy/cylinders/lagranger_bounder.py
```diff
@@ -113,4 +113,7 @@ def finalize(self):
         '''
         self.final_bound = self._update_weights_and_solve(self.A_iter)
         self.bound = self.final_bound
+        if self.opt.extensions is not None and \
+            hasattr(self.opt.extobject, 'post_everything'):
```
DLWoodruff, line 117: Ben's suggested that we *not* do conditional exceptions and instead add a pass to the base class for extensions when we add a new extension callout.

garg02: Oh I ended up doing both. I'll remove the conditional exceptions

DLWoodruff (approved):

PR #337, opened 2023-09-25
ph outer bounder

ulysse-n:
Outer bounder using PH with smaller rhos to compute a lower bound.
There are different ways of selecting rho values:
- by default (rescaling of hub value)
- manually
- use a json file with scaling factors
- use gradient-based rho (not tested)


DLWoodruff:
I want to merge changes to the rho_setter PR before merging this.

DLWoodruff (approved):

DLWoodruff:
Adding a new outer (lower) bounder spoke

PR #339, opened 2023-09-27
Nonant name validation

bknueven:
TODOs:
- [ ] Make test for bad farmer, both with local and global check
- [ ] See how expensive this check is


bknueven (commented):
Maybe we can use this module to avoid some unnecessary copying:

https://docs.python.org/3/library/struct.html

PR #346, opened 2023-11-02
Some spoke extensions

### LINKED ISSUE #345: xhatters and extensions need work
- we need a test extension that just asserts that it is called
- multi-extension seems to have some extra specification in options that needs to be documented
- xhatters need to allow for extensions (and that needs to be tested)
- the xhatter code in cfg vanilla should be factored

DLWoodruff:
Closes #345 
Allowed for extensions in many spokes
Some standardization in cfg_vanilla
Added a little doc concerning multiple extensions

### DIFF
--- a/mpisppy/spopt.py
+++ b/mpisppy/spopt.py
@@ -6,6 +6,7 @@ import logging
 import time
 import math
 import inspect
+import traceback
 
 import numpy as np
 from mpisppy import MPI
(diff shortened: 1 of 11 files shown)


mpisppy/spopt.py
```diff
@@ -6,12 +6,13 @@
 import time
 import math
 import inspect
+import traceback
```
bknueven, line 9:
```suggestion
```

bknueven (approved):

PR #347, opened 2023-11-14
random sleep with license retry

bknueven:
Not sure this is a good idea -- but usually a collection of ranks fails to acquire a license all at once, and then they all wait the same amount of time before re-trying.

This attempts to put some stochasticity into the retry time. Example output:
Screenshot 2023-11-14 at 3 02 06 PM

### DIFF
--- a/mpisppy/spopt.py
+++ b/mpisppy/spopt.py
@@ -6,6 +6,7 @@ import logging
 import time
 import math
 import inspect
+import random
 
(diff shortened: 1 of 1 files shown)


DLWoodruff (approved):
This is maybe not the very best thing, but surely better (and faster) than two seconds hard-wired.

jeanpaulwatson (approved):

PR #349, opened 2023-11-17
Termination callback

bknueven:
Digging up some code I wrote some time ago -- needs integration, but seems to work as-is.

In particular, now that calling `got_kill_signal` involves a few MPI operations (instead of just reading a location in local memory), we need to think carefully about how often we are polling that within a MIP solve.


mpisppy/utils/callbacks/termination/tests/test_termination.py
```diff
@@ -0,0 +1,99 @@
...
+@pytest.mark.skipif(
+    not SolverFactory("cplex_persistent").available(),
+    reason="cplex_persistent not available",
+)
+def test_cplex_termination_callback():
+    st = time.time()
+    cplextest = CPLEXTermination()
+    try:
+        cplextest.solve()
+    except ValueError:
```
jeanpaulwatson, line 65: This seems weird - might be worth documenting / exploring what happens.

jeanpaulwatson (approved):

jeanpaulwatson (approved):

PR #353, opened 2023-11-26
UC example quartz script and baseline updates

jeanpaulwatson:
To catch up with (1) opt-in instead of opt-out spoke inclusion arguments for uc_cylinders.py and (2) some recent speed improvements. 

Does not include cross-scenario cut variants. Those will come later, once some bugs/issues are sorted out.


DLWoodruff (approved):
Since these are baselines, maybe you should change the directory name from "quartz" to "quartz_Nov2023"

PR #354, opened 2023-11-27
add a denouement call for all xhatters

DLWoodruff:
I'm not sure this is a good idea (but we have a use case), and if it is, I'm not sure it is the best place in the class hierarchy


mpisppy/cylinders/spoke.py
```diff
@@ -350,6 +350,14 @@ def finalize(self):
             for var, value in s._mpisppy_data.best_solution_cache.items():
                 var.set_value(value, skip_validation=True)
 
+        # xhatters that don't want the denouement should remove it from their cylinder
+        #  (e.g. when they call vanilla)
+        # TBD: should this be higher in the class hierarchy?
```
bknueven, line 355: Hmm ... seems like a can of worms. Some spokes have algorithms, which call the below code automatically.

This is probably worth sorting out for consistency.

PR #358, opened 2023-12-21
Shufflefix

### LINKED ISSUE #357: XhatShuffleInnerBound keeps retrying the same solution
It seems like `XhatShuffleInnerBound` keeps retrying the same solution again and again until a new set of solutions is provided by the hub. If I'm right, I think the culprit might be that the nonanticipatory variables are not reset after trying the solution, so when the "next" solution is loaded, it just reloads the variable values that were just set and are still in the scenario.

https://github.com/Pyomo/mpi-sppy/blob/34bd37cbc5a84997c4baf3cc2806d89a521c71fe/mpisppy/cylinders/xhatshufflelooper_bounder.py#L69-L74

I think that changing `restore_nonants` from `False` to `True` in the call to `_try_one` will fix this.

A second observation, which maybe should be a separate issue, is that it seems like, if the hub takes long enough before giving a new solution, `XhatShuffleInnerBound` keeps looping over the same scenarios, even once it has already tried them all. I realize that maybe others don't consider this a problem if the assumption is that the number of scenarios is large compared to the solve time of the hub. I don't really understand what `XhatShuffleInnerBound` is doing with `begin_epoch`, `_iter_scen`, and the `nodescen_dict`, but it seems like the object is keeping track of tried scenarios in `_scenarios_this_epoch` such that it could avoid retrying scenarios that had already been tried. I have an idea of how this might be done, and I'd be glad to submit a PR.

DLWoodruff:
Partially fix for #357 
Xhatters that loop need to restore nonants during the loop, added a test to check this for xhatshufflelooper

### DIFF
--- a/mpisppy/cylinders/xhatlooper_bounder.py
+++ b/mpisppy/cylinders/xhatlooper_bounder.py
@@ -70,7 +70,7 @@ class XhatLooperInnerBound(spoke.InnerBoundNonantSpoke):
 
                 self.opt._put_nonant_cache(self.localnonants)
                 self.opt._restore_nonants()
-                upperbound, srcsname = xhatter.xhat_looper(scen_limit=scen_limit, restore_nonants=False)
+                upperbound, srcsname = xhatter.xhat_looper(scen_limit=scen_limit, restore_nonants=True)
 
                 # send a bound to the opt companion
                 self.update_if_improving(upperbound)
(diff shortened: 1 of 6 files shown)


bknueven (approved):
Code looks good -- I am not likely to get a chance to test this locally for a while, so I say we merge.

PR #368, opened 2024-02-13
Updates to proximal approximation

bknueven:
The current proximal approximation approach does not allow you to start with no cuts.

This PR would update the proximal approximation code such that:
1. Allows for starting with an empty set of cuts
2. Makes starting with the empty set of cuts the new default
3. Adds two cuts per pass: the "nearest point" cut (i.e., at x* = argmin [( x - \hat{x} )^2 + ( x^2 - \hat{y} )^2] and the "straightforward projection cut" (i.e., at x* = \hat{x}). Before we just added the "nearest point" cut.

### DIFF
--- a/mpisppy/phbase.py
+++ b/mpisppy/phbase.py
@@ -12,7 +12,7 @@ import mpisppy.utils.sputils as sputils
 import mpisppy.utils.listener_util.listener_util as listener_util
 import mpisppy.spopt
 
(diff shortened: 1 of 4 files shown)


DLWoodruff (approved):
This seems valuable.

bknueven:
After some more extensive testing, it isn't clear this is any better than the current implementation. Closing.

PR #372, opened 2024-03-06
Centralized subproblem creation

bknueven:
I want to add a new feature to `SPOpt`, but the way we create subproblems is making it awkward.

Unconditionally saves the scenario objective functions before bundle / subproblem creation.


DLWoodruff (approved):
This looks like a very sensible factoring.  I can't remember what the issue was with xhat_eval, but apparently moving the creation all the way up to __init__ takes care of it.

PR #373, opened 2024-03-11
Presolve

bknueven:
This PR adds "Distributed Subproblem Presolve".

This functionality is available for all Hub and Spoke algorithms which inherit from SPBase. It can be enabled by passing `presolve=True` into the constructor, and is available with the `--presolve` command line option.

Leveraging the existing feasibility-based bounds tightening (FBBT) available in Pyomo, this presolver will tighten the bounds on all variables, including the non-anticipative variables. If the non-anticipative variables have different bounds, the bounds among the non-anticipative variables will be synchronized to utilize the tightest available bound.

In its current state, the user might opt-in to presolve for two reasons:
1. For problems without relatively complete recourse, utilizing the tighter bounds on the non-anticipative variables and speed convergence and improve primal and dual bounds. In rare cases it might also detect infeasibility.
2. For problems where a “fixer” extension or spoke is used, determining tight bounds on the non-anticipative variables may improve the fixer’s performance.

The PR is structured in such a way that we can add additional presolve capability to mpi-sppy without too much effort.


bknueven:
TODO:
- [ ] Option-drive
- [ ] rst file (few sentence documentation)

bknueven:
@michaelbynum if you have time and interest in reviewing this I would appreciate any feedback you might have.

michaelbynum:
Very cool!! If I understand correctly, this is performing FBBT on each subproblem in parallel, communicating to update the non-ant bounds, and repeating? I think that is what you said above in different words?

mpisppy/opt/presolve.py
```diff
@@ -0,0 +1,247 @@
...
+
+        while True:
+            big_iters = 0.0
+            for k, it in self.subproblem_tighteners.items():
+                n_iters = it.perform_fbbt(self.opt.local_subproblems[k])
+                # get the number of constraints after we do
+                # FBBT so we get any updates on the subproblem
+                big_iters = max(big_iters, n_iters / len(it._cmodel.constraints))
+
+            update_this_pass = big_iters > 1.0
```
michaelbynum, line 101: This is saying, if any variables were tightened, keep going. Otherwise break. Right?

mpisppy/opt/presolve.py
```diff
@@ -0,0 +1,247 @@
...
+                    op=MPI.MIN,
+                )
+                if same_nonant_bounds:
+                    same_nonant_bounds = np.allclose(
+                        global_upper_bounds[ndn], local_bounds
+                    )
+
+            # At this point, we've either proved that
+            # there are tighter bounds or not.
+            # If not, we can quit
```
michaelbynum, line 190: I'm not sure this is precisely true, although I think I agree with the conclusion. It really depends on the stopping criteria for FBBT (improvement_tol and max_iter). It is technically possible to run FBBT twice and have the second iteration do something useful, even without any changes to the model in between. This is most likely when max_iter is small. Maybe just update the comment to say something like "the nonant bounds did not change, so it is unlikely that more FBBT will do anything"???

michaelbynum (approved):
Looks great! Just a few minor comments.

mpisppy/opt/presolve.py
michaelbynum: The default `max_iter` is relatively large, so this shouldn't be a problem. However, if you ever allow users to configure the `IntervalTightener`s for the subproblems, it could be an issue.

jeanpaulwatson (approved):
Conducted verbal/interactive feedback. Version does not break existing functionality, and works well on at least the netdes instance.

PR #377, opened 2024-04-02
Remove final Lagrangian solve

bknueven:
This can take a significant amount of time, such that finalization is waiting on the Lagrangian bounder.

The benefit seems marginal at best -- if we've already decided to stop, let's just wrap it up.


DLWoodruff (approved):
There should be an option to allow for this "final pass" because when PH is converging really quickly, this could be the only pass.

jeanpaulwatson (approved):

PR #381, opened 2024-04-16
try adding an init to the hydro test directory so pyotracker passes

DLWoodruff:


DLWoodruff:
this is a trivial change, so I will just merge it

PR #391, opened 2024-04-24
add subgradient option to lagrangian

bknueven:
Not sure this is a good idea -- you may miss incoming weights. Opening this draft PR for discussion.


DLWoodruff (commented):
Interesting idea.  Sort of a hybrid.  On the one hand, its behavior would be "erratic" in some sense because it would depend on hub solve times. On the other hand, it seems like a harmless option, particularly if it is only documented in the "secret menu" rst file.

PR #393, opened 2024-05-07
Dependency cleanup

bknueven:
Running `afew.py` on a new system today and a few dependencies surprised me.

I have relegated `matplotlib` to only be imported when a plotting function is called. I have also added a hard dependency on pandas as several built-in modules use it.

That said, I also get the following message on my system, which is part of the reason I hate having pandas as a dependency:
```
/Users/bknueven/Software/mpi-sppy/mpisppy/extensions/phtracker.py:63: FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
  self.df = pd.concat([self.df, new_dict], ignore_index=True)
```

### DIFF
--- a/mpisppy/convergers/primal_dual_converger.py
+++ b/mpisppy/convergers/primal_dual_converger.py
@@ -1,7 +1,6 @@
 import numpy as np
 import os
 import pandas as pd
-import matplotlib.pyplot as plt
 import mpisppy.convergers.converger
 from mpisppy import MPI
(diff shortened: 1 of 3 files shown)


DLWoodruff:
We just merged a fix to the pandas deprecation warning; however, maybe we should also work to make pandas not a hard requirement.

PR #403, opened 2024-06-12
trying to get a badge for pyotracker

DLWoodruff:


DLWoodruff:
Worth a try; at the very least it won't say "failing" and will provide a link to the test

PR #410, opened 2024-08-16
Lagranger: Be sure we don't set nonants to `nan`

bknueven:
If Lagranger is faster than the hub to iteration 1, the nonants from the hub will be `nan`, which can cause the `W`'s to be `nan`, and then the user gets an error from the solver interface.

The PR would fix this bug.

### DIFF
--- a/mpisppy/cylinders/lagranger_bounder.py
+++ b/mpisppy/cylinders/lagranger_bounder.py
@@ -6,6 +6,7 @@
 import time
 import json
 import csv
+import numpy as np
 import mpisppy.cylinders.spoke
 from mpisppy.cylinders.lagrangian_bounder import _LagrangianMixin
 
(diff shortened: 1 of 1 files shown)


DLWoodruff (approved):
Please merge this (even though I am sort of disowning Lagranger). Doesn't lagrangian need the same thing?
  Dave

bknueven:
> Please merge this (even though I am sort of disowning Lagranger). Doesn't lagrangian need the same thing? Dave

Well, this is weird. On the farmer instance this came up on, we no longer crash, but now I see two different behaviors depending on "who wins".

If we don't have `nan`s first, I see:
```
[    0.63] Iter.           Best Bound  Best Incumbent      Rel. Gap        Abs. Gap
[    0.63]     1  A      -115405.5556             inf           inf%             inf
[    0.64]     2  A      -112378.3951             inf           inf%             inf
[    0.65]     3         -112378.3951             inf           inf%             inf
[    0.66]     4  A      -111904.8677             inf           inf%             inf
[    0.68]     5  A      -110093.8156             inf           inf%             inf
[    0.69]     6  A      -109499.6274             inf           inf%             inf
[    0.70]     7  A      -108932.8877             inf           inf%             inf
[    0.71]     8         -108932.8877             inf           inf%             inf
[    0.72]     9         -108932.8877             inf           inf%             inf
[    0.73]    10         -108932.8877             inf           inf%             inf
[    0.73] Reached user-specified limit=10 on number of PH iterations
```

But if we do have `nan`s at the beginning, then I get:
```
[    0.62] Iter.           Best Bound  Best Incumbent      Rel. Gap        Abs. Gap
[    0.62]     1  A      -112378.3951             inf           inf%             inf
[    0.63]     2         -112378.3951             inf           inf%             inf
[    0.64]     3         -112378.3951             inf           inf%             inf
[    0.65]     4         -112378.3951             inf           inf%             inf
[    0.66]     5         -112378.3951             inf           inf%             inf
[    0.67]     6         -112378.3951             inf           inf%             inf
[    0.69]     7         -112378.3951             inf           inf%             inf
[    0.70]     8         -112378.3951             inf           inf%             inf
[    0.71]     9         -112378.3951             inf           inf%             inf
[    0.72]    10         -112378.3951             inf           inf%             inf
[    0.72] Reached user-specified limit=10 on number of PH iterations
```

bknueven:
On Lagrangian, it patiently waits for new W's, so this doesn't come up. Lagranger does not wait..

PR #413, opened 2024-08-22
Gradient-based rho integration

jeanpaulwatson:
We are not 100% happy with this, but capability now works on UC examples and works well. Re-factoring and code review of all modules is suggested in the near future.


examples/uc/gradient_uc_cylinders.py
bknueven, line 1: Should this be separate from `uc_cylinders.py`?

mpisppy/utils/config.py
```diff
@@ -653,67 +657,85 @@ def cross_scenario_cuts_args(self):
...
+
+        self.gradient_args()
+
+        self.add_to_config('grad_dynamic_primal_crit',
+                           description="Use dynamic gradient-based primal criterion for update",
+                           domain=bool,
+                           default=False)
+
+        self.add_to_config('grad_dynamic_dual_crit',
+                           description="Use dynamic gradient-based primal criterion for update",
```
bknueven, line 725:
```suggestion
                           description="Use dynamic gradient-based dual criterion for update",
```

bknueven (approved):

PR #425, opened 2024-09-18
Add `ruff` linter fixes and checks

bknueven:


mpisppy/utils/listener_util/demo_listener_util.py
```diff
@@ -234,7 +231,7 @@ def side_gig(synchro, msg = None):
 # Note: at this point the first reduce is the only reduce
 Lens = collections.OrderedDict({"FirstReduce": {"ROOT": 2+n_proc}})
 if rank == 0:
-    logging.debug("iters %d, sleep %f, seed %d".format((iters, sleep, seed)))
+    logging.debug("iters %d, sleep %f, seed %d".format())
```
DLWoodruff, line 234: This doesn't make sense to me. What will an empty format do?

DLWoodruff (approved):

mpisppy/utils/listener_util/demo_listener_util.py
bknueven: I think I fixed this in fb28cd50. The original code mishmashed two different ways of formatting strings in Python (there are obviously a few too many...).

PR #402, opened 2024-06-10
Agnostic

DLWoodruff:
With respect to the guest side, it is very preliminary. On the host side, the callouts seems to work well, but more features probably need callouts.

PR #426, opened 2024-09-18
Add formatter

bknueven:


bknueven:
We should actually merge #402 before this, as it will be a terrible re-basing exercise.

DLWoodruff (approved):
This is going to raise heck with every active PR and branch, but let's go for it

PR #429, opened 2024-09-20
Adding iter0_post_solver_creation callback to extensions object

jeanpaulwatson:
Motivated by the timed mipgap termination extension, which needs solver plugins to be available in order to install itself.


mpisppy/phbase.py
```diff
@@ -827,9 +826,12 @@ def Iter0(self):
                 stochastic program with the nonanticipativity constraints
                 removed.
         """
+
+        # JPW: the pre iter0 callback should arguably be called not from within
```
DLWoodruff, line 830: There are multiple pathways to iter0, so I would argue that this is the best place. But, in any event, this comment should probably go somewhere else (e.g. an issue)

jeanpaulwatson: Fair - corrected (and opened an issue).

DLWoodruff (approved):

PR #438, opened 2024-10-03
change default for cfg add to be complain

DLWoodruff:


mpisppy/utils/config.py
```diff
@@ -98,7 +98,7 @@ def add_and_assign(self, name, description, domain, default, value, complain=Fal
         """
         if name in self:
             if complain:
-                print(f"Duplicate {name} will not be added to self by add_and_assign {value}.")
+                print(f"Duplicate {name=} will not be added to cfg to assign {value=}")
```
bknueven, line 101: Should the "complain" option raise the exception below instead?

DLWoodruff: Probably. I'll make that change.

bknueven (approved):

PR #439, opened 2024-10-09
Lagrangian does not raise error if no solution is available

bknueven:
Lagrangian doesn't need a primal solution, just a lower bound. For MIPs, configuring the solver to only focus on the bound in the Lagrangian spoke is often beneficial.


mpisppy/cylinders/reduced_costs_spoke.py
DLWoodruff, line 1: I think line 87 should be need_solution=need_solution

DLWoodruff, line 1: I think line 87 should be need_solution=need_solution

DLWoodruff (approved):

mpisppy/cylinders/reduced_costs_spoke.py
bknueven: Yes, maybe. I've probably OOP-ed myself into a corner with this one, because the method which calls this is `LagrangianOuterBound.main`, which does not need the solution.

PR #440, opened 2024-10-10
Proper Bundles

DLWoodruff:
creates, uses, reads and writes proper bundles
Proper bundles of size 1 are supported
Also supports pickled scenarios


mpisppy/utils/config.py
```diff
@@ -951,6 +951,22 @@ def wxbar_read_write_args(self):
...
+        self.add_to_config('unpickle_bundles_dir',
+                            description="Read bundles from a dill pickle files in this dir; (default None)",
+                            domain=str,
+                            default=None)
+
+        self.add_to_config("scenarios_per_bundle",
+                            description="Used for `proper` bundles only (might also be used when reading pickled bundles) (default None)",
+                            domain=int,
+                            default=None)
+
```
bknueven, line 969:
```suggestion
    def proper_bundle_config(self):
        self.add_to_config('pickle_scenarios_dir',
                            description="Write scenarios to a dill pickle files in this dir (default None)",
                            domain=str,
                            default=None)

        self.add_to_config('unpickle_scenarios_dir',
                            description="Read scenarios from a dill pickle files in this dir; (default None)",
                            domain=str,
                            default=None)

        self.add_to_config("scenarios_per_bundle",
                            description="Number of scenarios in each bundle (default 1)",
                            domain=int,
                            default=1)

```
My last API suggestion. If we do this these config options probably need to be moved to a different method.

bknueven (approved):
I like it as-is, but if we think about this PR as both "scenario pickling capability" and "proper bundles" as separate concepts it is simpler, IMHO. Maybe for a follow-up.

bknueven:
Another high-level comment: I think we should move the proper bundling capability into SPBase.

PR #443, opened 2024-10-25
Fix bug in FWPH with Pyomo 6.8

### LINKED ISSUE #442: FWPH spoke does not work with non-persistent solvers
Error message:
```
AttributeError: 'ScalarConstraint' object has no attribute '_body'. Did you mean: 'body'?
```

We probably need to reflect this change in FWPH: https://github.com/Pyomo/pyomo/pull/3293/files#diff-abc64d73619e2aec8965fd7c89195ee22162f31c2ef7d3d1b9537311ef76868eL265-R267

bknueven:
Fixes #442. Adds additional runall runner for direct solvers.

### DIFF
--- a/.github/workflows/test_pr_and_main.yml
+++ b/.github/workflows/test_pr_and_main.yml
@@ -91,8 +91,8 @@ jobs:
           cd ./doc/src/
           make doctest
 
-  runall:
-    name: run_all.py
+  runall_persistent:
+    name: run_all.py persistent solver
     runs-on: ubuntu-latest
     timeout-minutes: 15
     # this takes the most time, so we'll start it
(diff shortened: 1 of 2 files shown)


bknueven:
@DLWoodruff I'm having trouble with APH when using xpress_direct. I don't really understand the need for `disable_pyomo_signal_handling`, and it seems like it weirdly fails for direct solvers but not persistent solvers.

A while back Pyomo changed the default behavior for this https://github.com/Pyomo/pyomo/pull/1014, so I'm not sure it's even needed anymore. But I know very little about APH...

DLWoodruff:
I also don't understand the need to disable signal handling in APH. It has something to do with the listener/worker thread architecture. I will try to fix it in another PR.

DLWoodruff (approved):
The run_all test is still running, but this looks good.

PR #447, opened 2024-11-01
Reduced cost rho

bknueven:
Rho setter using reduced costs from a reduced cost spoke


DLWoodruff (approved):
Currently, the only test for reduced costs is one run_all.py, which is better than nothing.  Maybe after this is merged, I can remember to add something to generic_tester.py

DLWoodruff (approved):
Much cleaner and clearer

PR #464, opened 2024-12-02
Prox approx part 3

bknueven:
- Re-implement cut manager based on tolerance
- Add cuts based on W
- Add safety cuts if W == 0 and xbar == xhat
- Don't bother if the variable is fixed
- Skip cut if it is outside the variable bounds

### DIFF
--- a/mpisppy/utils/sputils.py
+++ b/mpisppy/utils/sputils.py
@@ -984,7 +984,8 @@ def nonant_cost_coeffs(s):
         if id(var) in s._mpisppy_data.varid_to_nonant_index:
             raise RuntimeError(
                 "Found nonlinear variables in the objective function. "
-                f"Variable {var} has nonlinear interactions in the objective funtion"
+                f"Variable {var} has nonlinear interactions in the objective funtion. "
+                "Consider using gradient-based rho."
             )
     return cost_coefs
 
--- a/mpisppy/utils/sputils.py
+++ b/mpisppy/utils/sputils.py
@@ -983,7 +983,7 @@ def nonant_cost_coeffs(s):
     for var in repn.nonlinear_vars:
         if id(var) in s._mpisppy_data.varid_to_nonant_index:
             raise RuntimeError(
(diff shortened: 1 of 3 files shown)


mpisppy/utils/sputils.py
```diff
@@ -984,7 +984,8 @@ def nonant_cost_coeffs(s):
         if id(var) in s._mpisppy_data.varid_to_nonant_index:
             raise RuntimeError(
                 "Found nonlinear variables in the objective function. "
```
DLWoodruff, line 986: Change "Found nonlinear..." to "A call to nonant_cost_coefficient found nonlinear" (even though the user could figure that out from the stack trace, I don't want to give the impression that nonlinear is not OK in general")

DLWoodruff (approved):

bknueven:
Not really sure if this PR or #467 should be merged instead.

PR #467 takes a more systematic approach, and hence adds more cuts. This can slow subproblem solves, but the iterations seem to be of higher quality due to the extra cuts.

DLWoodruff:
> Not really sure if this PR or #467 should be merged instead.
> 
> PR #467 takes a more systematic approach, and hence adds more cuts. This can slow subproblem solves, but the iterations seem to be of higher quality due to the extra cuts.

To state the obvious: we could have both and that might even be a good idea (but does require even more work)

mpisppy/utils/sputils.py
bknueven, line 986:
```suggestion
                "A call to nonant_cost_coefficient found nonlinear variables in the objective function. "
```

bknueven:
> > Not really sure if this PR or #467 should be merged instead.
> > PR #467 takes a more systematic approach, and hence adds more cuts. This can slow subproblem solves, but the iterations seem to be of higher quality due to the extra cuts.
> 
> To state the obvious: we could have both and that might even be a good idea (but does require even more work)

Yes. I am confident this is an improvement over `main`, so let's merge this. It is simpler, and then we can consider the additional changes in #467 separately.

PR #467, opened 2024-12-02
Prox approx 6

bknueven:


bknueven:
I think this should be merged, but I'm a bit concerned that there's an assumption in `add_initial_cuts_var` that `xbar == 0`.

bknueven:
> I think this should be merged, but I'm a bit concerned that there's an assumption in `add_initial_cuts_var` that `xbar == 0`.

Actually, the assumption is `xbar == xhat`, which is fine for this code because we only get to it if `| xbar - xhat | < tolerance`.

DLWoodruff (approved):

PR #470, opened 2024-12-15
Lp writer

DLWoodruff:
This extension writes an lp file with the model and json file with (a) list(s) of
scenario tree node names and nonanticaptive variables for each scenario before
the iteration zero solve of PH or APH.


mpisppy/extensions/scenario_lpfiles.py
```diff
@@ -0,0 +1,40 @@
...
+""" Extension to write an lp file for each scenario and a json file for
+the nonant structure for each scenario (yes, for two stage problems this
+json file will be the same for all scenarios.)
+"""
+
+import json
+import mpisppy.extensions.extension
+
+
+def lpize(varname):
```
bknueven, line 18: Is there a function already in pyomo we can just import to do this?

bknueven (approved):

mpisppy/extensions/scenario_lpfiles.py
DLWoodruff: good idea

PR #474, opened 2024-12-28
Proper multistage bundles

DLWoodruff:
The bundles must consume the same number of entire second stage nodes, which results in a two-stage problem.


bknueven:
Is this tested?

bknueven (approved):

PR #476, opened 2025-01-10
Move to single common MPI Window between Hub and Spokes

jmaack24:
Change the hub and spoke usage of MPI windows to use a single shared window between the hub and all spokes. This supports spokes pulling values from other spokes as well as the hub.


mpisppy/cylinders/spoke.py
```diff
@@ -187,22 +174,20 @@ def bound(self):
...
-        self.spoke_to_hub(self._bound)
+        self.spoke_to_hub(self._bound, self.bound_type())
+        return
 
     @property
     def hub_inner_bound(self):
         """Returns the local copy of the inner bound from the hub"""
-        return self._locals[-2]
+        # NOTE: This should be the same as _hub_bounds[1]
+        return self._hub_bounds[-2]
```
bknueven, line 184: Maybe should just make this `_hub_bounds[1]` then, since the code supports this more straightforward indexing now

mpisppy/cylinders/spoke.py
```diff
@@ -187,22 +174,20 @@ def bound(self):
...
     def hub_outer_bound(self):
         """Returns the local copy of the outer bound from the hub"""
-        return self._locals[-3]
-
-    def _got_kill_signal(self):
-        """Looks for the kill signal and returns True if sent"""
-        self._new_locals = self.spoke_from_hub(self._locals)
-        return self.remote_write_id == -1
+        # NOTE: This should be the same as _hub_bounds[0]
+        return self._hub_bounds[-3]
```
bknueven, line 190: Similar here

mpisppy/cylinders/spoke.py
```diff
@@ -216,27 +201,24 @@ class _BoundNonantLenSpoke(_BoundSpoke):
         want something of len nonants from OPT
     """
 
-    def make_windows(self):
-        """ Makes the bound window and with a remote buffer long enough
-            to hold an array as long as the nonants.
+    @abc.abstractmethod
+    def nonant_len_type(self) -> Field:
+        # TODO: Make this a static method?
+        pass
```
bknueven, line 207: Yeah. It could also be a class attribute which defaults to `None`.

bknueven: Doesn't need to be done for this PR.

mpisppy/cylinders/spwindow.py
```diff
@@ -0,0 +1,131 @@
...
+    SHUTDOWN=-1000
+    NONANT=1
+    DUALS=2
+    OBJECTIVE_BOUNDS=100 # Both inner and outer bounds from the hub. Layout: [OUTER INNER ID]
+    OBJECTIVE_INNER_BOUND=101
+    OBJECTIVE_OUTER_BOUND=102
+    EXPECTED_REDUCED_COST=200
+    SCENARIO_REDUCED_COST=201
+    CROSS_SCENARIO_CUT=300
+    NONANT_COEFS=400
```
bknueven, line 27: Because this is used for sending `eta`, which is the recourse cost for scenario s_1 as observed by scenario s_2, it seem like it should be named something else, like `CROSS_SCENARIO_COST` or something similar.

bknueven (commented):
A few minor comments. The only thing I really want changed before merging however renaming `NONANT_COEFFS`. I also want to do a few tests on Kestrel to ensure there's no unanticipated performance degradation (I expect this will perform better, FWIW).

mpisppy/cylinders/spwindow.py
jmaack24: Renamed NONANT_COEFS to CROSS_SCENARIO_COST as requested.

jmaack24:
> We will merge this soon. On the dev call we decided we wanted to stress test this to ensure there's no regression on clean, non-hanging termination.

Sounds good. Let me know if you need any help with stress testing.

bknueven (approved):
I ran 10 iterations of scalable farmer 300 times on my laptop with number of ranks 1-2 times the number of cores on my laptop, while doing my other work.

Not a single instance hung due to a race condition at termination.

bknueven:
> > I ran 10 iterations of scalable farmer 300 times on my laptop with number of ranks 1-2 times the number of cores on my laptop, while doing my other work.
> > Not a single instance hung due to a race condition at termination.
> 
> I'm going to run for 1000 iterations several times with 3-crop farmer and lots of scenarios / ranks to verify before we merge.

I ran several 1000-iteration problems on my laptop in several unusual configurations without termination issues.

PR #477, opened 2025-01-10
Make sure prox_approx does not loop forever

bknueven:
Drive by: tee all ranks because we say we're tee-ing all ranks. The caller can control who has `tee=True`


bknueven:
Replaced by #480.

PR #481, opened 2025-01-23
fixing xhatters

bknueven:
Follow up from #461.

The changes in #461 were needed to ensure the solution written was nonanticipative. However, as a side effect, it made the xhatters less effective. Because `XhatBase._try_one` did not restore the nonants, when the "next" scenario was called it would have the previous solution loaded, and not the PH iteration loaded. Therefore the incumbent finders were only producing at most one solution per PH iteration.

This PR fixes the same bug as #461, but does so my moving the best solution found tracking code into `SPBase`. Then the Xhatters can update and cache the solution immediately after it is found, and then finally restoring the nonants for the next try. This also makes it much less likely that solutions found will be accidentally dropped because of a user configuration error.

### DIFF
--- a/mpisppy/cylinders/xhatshufflelooper_bounder.py
+++ b/mpisppy/cylinders/xhatshufflelooper_bounder.py
@@ -83,7 +83,7 @@ class XhatShuffleInnerBound(spoke.InnerBoundNonantSpoke):
             return False
         _vb(f"    Feasible {snamedict}, obj: {obj}")
 
-        # the xhatter updates the cache in the opt object for us
+        # XhatBase._try_one updates the solution cache in the opt object for us
         update = self.update_if_improving(obj, update_cache=False)
         logger.debug(f'   bottom of try_scenario_dict on rank {self.global_rank}')
         return update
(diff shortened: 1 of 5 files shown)


mpisppy/cylinders/xhatshufflelooper_bounder.py
```diff
@@ -83,7 +83,8 @@ def _vb(msg):
             return False
         _vb(f"    Feasible {snamedict}, obj: {obj}")
 
-        update = self.update_if_improving(obj)
+        # the xhatter updates the cache in the opt object for us
```
DLWoodruff, line 86: I don't understand this comment

bknueven: It happens [here](https://github.com/Pyomo/mpi-sppy/blob/a9c522173ca37459d2b311de5caf2ed08c5b38ad/mpisppy/extensions/xhatbase.py#L205) and [here](https://github.com/Pyomo/mpi-sppy/blob/a9c522173ca37459d2b311de5caf2ed08c5b38ad/mpisppy/extensions/xhatbase.py#L232).

On critical thing is that `SPBase` now tracks the best solution found so far, as opposed to the incumbent spoke.

bknueven: Basically, you want to cache the solution *before* the xhatter restores the nonants.

DLWoodruff (approved):

PR #486, opened 2025-01-30
less verbose presolve by default

bknueven:


bknueven:
I think I'm not getting the Spoke name printed because the `spcomm` object is attached *after* presolve is called.

jeanpaulwatson (approved):

PR #488, opened 2025-02-07
Xbar-W reader writer options processing re-work

jeanpaulwatson:
Intended as an exemplar for improving options processing and reading in mpisppy. 

Objective is to move options definition and processing into "leaf" classes, to improve encapsulation. 

This PR only does this for XBar-W reader/writer extensions, as a concise place to start.

Also actually makes use of these extensions in generic_cylinders.py - previously the options were specified, but never used.

### DIFF
--- a/mpisppy/utils/config.py
+++ b/mpisppy/utils/config.py
@@ -959,11 +959,11 @@ class Config(pyofig.ConfigDict):
                             default=0)
 
     def wxbar_read_write_args(self):
-        import mpisppy.utils.wxbarreader
+        import mpisppy.utils.wxbarreader as wxbarreader
         print("DEPRECATION WARNING: wxbar_read_write_args should be moved to wxbarreader.add_options_to_config")
         wxbarreader.add_options_to_config(self)
 
-        import mpisppy.utils.wxbarwriter
+        import mpisppy.utils.wxbarwriter as wxbarwriter
         print("DEPRECATION WARNING: wxbar_read_write_args should be moved to wxbarwriter.add_options_to_config")
         wxbarwriter.add_options_to_config(self)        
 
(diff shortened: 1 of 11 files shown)


mpisppy/utils/config.py
```diff
@@ -958,32 +958,12 @@ def tracking_args(self):
                             domain=int,
                             default=0)
 
-
     def wxbar_read_write_args(self):
```
bknueven, line 961: Is this idea that this method would eventually go away so that `generic_cylinders` or any other clients directly call the `add_options_to_config` on the module?

mpisppy/utils/wxbarreader.py
```diff
@@ -39,18 +39,42 @@
...
+
 class WXBarReader(mpisppy.extensions.extension.Extension):
     """ Extension class for reading W values
     """
     def __init__(self, ph):
 
+        assert 'cfg' in ph.options
+        self.cfg = ph.options['cfg']
+        if self.cfg.get("W_and_xbar_reader") is None or not self.cfg.W_and_xbar_reader:
+            self.not_active = True
```
bknueven, line 69: Since this is new code, I'll just say I don't like negative `bool`s (i.e., the flag should be called `active`).

Not a reason to hold up this PR.

bknueven (approved):
Looks good. I have a question and a nit, but I'm okay to merge it now.

mpisppy/utils/config.py
DLWoodruff: Yes.

mpisppy/utils/wxbarreader.py
DLWoodruff: I will make the change.

PR #498, opened 2025-03-25
added a call to an initialize function to generic_cylinders per a sug…

DLWoodruff:
…gestion by Tomas

### DIFF
--- a/doc/src/generic_cylinders.rst
+++ b/doc/src/generic_cylinders.rst
@@ -42,3 +42,34 @@ ignored.
    probably needed on the command line. Consistency with the files in the
    pickle directory might not be checked by the program.
 
+A Class in the module
+---------------------
+
+If you want to have a class in the module to help create return values
+for functions, you will need to have an additional helper file called
+``initialize(cfg)``. It is called by ``generic_cylinders.py``
+after cfg is populated and can
+be used to create a class. Note that the function ``inparser_adder`` cannot
+make use of the class because that function is called before ``initialize``.
+Here is some sudo-code for this idea:
+
+.. code_block:: python
+
+    my_object = None
+
+    def inparser_adder(cfg):
+        cfg.add_to_config('json_file_with_model_params',
+                      ...)
+
+    def initalize(cfg):
+        global my_object
+        my_object = MyClass(cfg)
+
+    def scenario_names_creator(num_scens,...):
+        return my_object.scenario_names_creator(num_scens,...)
+
+    def scenario_creator(scen_name, ...):
+        return my_object.scenario_creator(scen_name, ...)
+
+def scenario_denouement(rank, ...):
+    return my_object.scenario_denoument(rank, ...)   
--- a/mpisppy/generic_cylinders.py
+++ b/mpisppy/generic_cylinders.py
@@ -544,9 +544,13 @@ if __name__ == "__main__":
         fname = os.path.basename(model_fname)
         sys.path.append(dpath)
         module = importlib.import_module(fname)
-    
+
     cfg = _parse_args(module)
 
+    # the initialize function is particularly useful for a class in the module
+    if hasattr(module, "initialize"):
+        module.initialize(cfg)
+    
     bundle_wrapper = None  # the default
     if _proper_bundles(cfg):
         # TBD: remove the need for dill if you are not reading or writing
--- a/examples/sizes/sizes.py
+++ b/examples/sizes/sizes.py
@@ -82,6 +82,11 @@ def sample_tree_scen_creator(sname, stage, sample_branching_factors, seed,
     sca["num_scens"] = sample_branching_factors[0]  # two-stage problem
     return scenario_creator(sname, **sca)
 
+def initialize(cfg):
+    print("initialize was successfully called."
+          " BTW: it does nothing for the sizes problem,"
+          " it is just here for testing purposes.")
+
 ######## end helper functions #########
 
 ########## a customized rho setter #############


bknueven (commented):
I'm a little confused by this. If a user wanted to use a class instead of a module, couldn't we enable that fairly easily? E.g.,

```python
class MyClassForMpiSppy:

    def __init__(self, cfg):
        # anything sensible
        pass

    def scenario_creator(self, scenario_name, **kwargs):
        ## build the scenario
        return scenario
         

# Maybe here, maybe somewhere else. Could have a different name.
# If we see this name in the module, we construct this class instead and
# use it in place of `module` in generic_cylinders.py.
mpisppy_class = MyClassForMpiSppy

# Or, we could require the user to provide a function which returns an
# instance of this class
def get_mpisppy_data(cfg):
    return MyClassForMpiSppy(cfg)
```

DLWoodruff:
I like your second idea best. I would change the function name to be
get_function_class

mpisppy/generic_cylinders.py
```diff
@@ -544,9 +552,13 @@ def _proper_bundles(cfg):
         fname = os.path.basename(model_fname)
         sys.path.append(dpath)
         module = importlib.import_module(fname)
-    
+
     cfg = _parse_args(module)
 
+    # Perhaps use an object as the so-called module.
+    if hasattr(module, "get_function_object"):
+        module = module.module_get_function_object(cfg)
```
bknueven, line 560:
```suggestion
        module = module.get_function_object(cfg)
```
`function` and `object` are very generic words. I still like `get_mpisppy_methods` or something similar.

mpisppy/generic_cylinders.py
```diff
@@ -493,6 +496,11 @@ def _do_EF(module, cfg, scenario_creator, scenario_creator_kwargs, scenario_deno
         sputils.ef_ROOT_nonants_npy_serializer(ef, f'{cfg.solution_base_name}.npy')
         sputils.write_ef_tree_solution(ef,f'{cfg.solution_base_name}_soldir')
         global_toc("Wrote EF solution data.")
+        # callout to special solution writer if it exists
+        if hasattr(module, 'custom_writer'):
+            module.custom_writer(ef, cfg)
```
tvalenciaz, line 501:
```suggestion
    # callout to special solution writer if it exists
    if hasattr(module, 'custom_writer'):
        module.custom_writer(ef, cfg)    
```
Placing it outside the if block, so that it is called independently of the value of `cfg.solution_base_name`, like for `_do_decomp()`

doc/src/generic_cylinders.rst
```diff
@@ -42,3 +42,41 @@ ignored.
...
+
+Assuming the first formal parameter in your function is assigned
+to variable named wheel in the non-ef case, your function probably should
+include something along the lines of this:
+
+.. code_block:: python
+
+   if wheel.spcomm.opt.cylinder_rank == 0:
+
+so that you avoid writing the output from every rank.
```
tvalenciaz, line 82: I would add something along the lines of:

Note this verification is automatically performed by WheelSpinner if you call your custom writer functions through 
`wheel.write_first_stage_solution(solution_file_name, first_stage_solution_writer=my_first_stage_writer)` and `wheel.write_tree_solution(solution_dir_name, scenario_tree_solution_writer=my_tree_solution_writer)`

mpisppy/generic_cylinders.py
DLWoodruff: How about get_mpisppy_helper_object?

bknueven: I could live with it :-)

DLWoodruff: done.

doc/src/generic_cylinders.rst
DLWoodruff: done.

DLWoodruff:
This still needs a test/example.

doc/src/generic_cylinders.rst
```diff
@@ -42,3 +42,41 @@ ignored.
...
+Note this verification is automatically performed by WheelSpinner if you call your custom writer functions through
+``wheel.write_first_stage_solution(solution_file_name, first_stage_solution_writer=my_first_stage_writer)`` and ``wheel.write_tree_solution(solution_dir_name, scenario_tree_solution_writer=my_tree_solution_writer)``.
```
bknueven, line 82: Should we just be handling this for the user, e.g., pass the function they provide into `write_first_stage_solution` or `write_tree_solution`? These methods also determine *which* cylinder has the optimal solution to write, so it's a bit more subtle than checking the cylinder rank. Plus, if you need the tree solution, you have to write from every rank.

doc/src/generic_cylinders.rst
```diff
@@ -42,3 +42,41 @@ ignored.
...
+A Class in the module
+---------------------
+
+If you want to have a class in the module to provide helper functions,
+your module still needs to have an ``inparse_adder`` function and the module will need
+to have a function called ``get_mpisppy_helper_object(cfg)`` that returns
+the object.  It is called by ``generic_cylinders.py`` after cfg is
+populated and can be used to create a class. Note that the function
+``inparser_adder`` cannot make use of the class because that function
+is called before ``get_function_object``.
```
bknueven, line 54:
```suggestion
is called before ``get_mpisppy_helper_object``.
```

doc/src/generic_cylinders.rst
DLWoodruff: Good point.

tvalenciaz: Agreed, that is probably safer and does not really restrict what the user can do; they can just put it inside the corresponding custom writer.

PR #504, opened 2025-05-01
adding remaining features from the CEP fork of mpisppy

tvalenciaz:


mpisppy/spopt.py
```diff
@@ -166,6 +167,10 @@ def _vb(msg):
             _vb("Using sub-problem solver options="
                 + str(solver_options))
             for option_key,option_value in solver_options.items():
+                if option_key=='LogFile':
```
bknueven, line 170: need to check for gurobi and/or make this solver-specific

mpisppy/spopt.py
```diff
@@ -166,6 +167,10 @@ def _vb(msg):
...
+                    lfn,lfn_e = os.path.splitext(option_value)
+                    lfn+=('_'+str(self.cylinder_rank))
```
bknueven, line 172: Shouldn't this be the global rank?

bknueven, line 172:
```suggestion
                    lfn += f"_{self.__class__.__name__}_{self.cylinder_rank}"
```

PR #512, opened 2025-05-05
Enhancements for proper bundles

bknueven:
- Support non-uniform probabilities in proper bundles
- Support surrogate non-anticipative variables in proper bundles
- Proper bundles now doesn't build the first scenario twice
- Add `surrogate_nonant` option for SSLP when using generic_cylinders.py
- Correctly compute nonant cost coefficients when proper bundles are used

### DIFF
--- a/doc/src/properbundles.rst
+++ b/doc/src/properbundles.rst
@@ -77,8 +77,7 @@ file ``aircond_cylinders.py`` in the aircond example directory
 provides an example.  The latter part of the ``allways.bash`` script
 demonstrates how to run it.
 
-There is support for multi-stage bundles in mpi-sppy, but the scenario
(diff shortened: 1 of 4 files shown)


DLWoodruff (commented):
If the original scenarios have probabilities that are not uniform, their sum needs to be the bundle probability and I don't see where that sum is computed.

bknueven:
> If the original scenarios have probabilities that are not uniform, their sum needs to be the bundle probability and I don't see where that sum is computed.

We were already computing it here, I believe:
https://github.com/Pyomo/mpi-sppy/blob/3af336fb1f891b2226c155b133d0be6117310d30/mpisppy/utils/sputils.py#L342

DLWoodruff (approved):

PR #515, opened 2025-05-12
added missing __init__.py to mpisppy/utils/kkt

bwpriest:
`mpisspy/utils/kkt/` is missing an `__init__.py`, which I assume is an oversight as all other subpackages appear to have one. This causes import errors when installing mpisppy as a repo dependency in another project's `pyproject.toml`. Adding a blank `__init__.py` should fix the problem by allowing `setuptools` to correctly identify the `mpisppy/utils/kkt/` directory as a subpackage.


bknueven:
Thanks!

Issue #516, opened 2025-05-12
Update build system to support PEP 517


bwpriest:
Currently, when including `mpi-sppy` as a dependency in another project's PEP 517-compliant `pyproject.toml`, newer versions of `pip` issue the following warning:

```
DEPRECATION: Building 'gridx-egret' using the legacy setup.py bdist_wheel mechanism, which 
will be removed in a future version. pip 25.3 will enforce this behaviour change. A possible 
replacement is to use the standardized build interface by setting the `--use-pep517` option, 
(possibly combined with `--no-build-isolation`), or adding a `pyproject.toml` file to the source
tree of 'gridx-egret'. Discussion can be found at https://github.com/pypa/pip/issues/6334
```

Additionally, the [docs](https://mpi-sppy.readthedocs.io/en/latest/quick_start.html#quick-start) for this project currently instruct to build from source using `python setup.py develop`, which is deprecated in `setuptools` (see this [note](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/) for details).

https://github.com/pypa/pip/issues/6334 and the PEP 517 [spec](https://peps.python.org/pep-0517/) recommend adding a `pyproject.toml` and to produce an editable build from source using `pip install -e ${REPO_ROOT}` to elegantly fix these issues. A `pyproject.toml` for this project (directly translating from the `setup.py`), will look something like:

```
###############################################################################
# mpi-sppy: MPI-based Stochastic Programming in PYthon
#
# Copyright (c) 2024, Lawrence Livermore National Security, LLC, Alliance for
# Sustainable Energy, LLC, The Regents of the University of California, et al.
# All rights reserved. Please see the files COPYRIGHT.md and LICENSE.md for
# full copyright and license information.
###############################################################################

[build-system]
[build-system]
requires = ["setuptools >= 68.0.0", "wheel"]
build-backend = "setuptools.build_meta"

[project]
name = "mpi-sppy"
version = "0.12.2.dev0"
authors = [
    {name = "David Woodruff", email = "dlwoodruff@ucdavis.edu" },
]
description = "mpi-sppy"
readme = "README.md"
license = { file = "LICENSE.md" }
requires-python = ">=3.9"
dependencies = [
    "sortedcollections",
    "numpy<2",
    "scipy",
    "pyomo>=6.4",
]

[project.urls]
homepage = "https://github.com/Pyomo/mpi-sppy"
source = "https://github.com/Pyomo/mpi-sppy"
[...]

PR #519, opened 2025-05-14
update build system

bwpriest:
Addresses issue #516 by replacing setup.py with a pyproject.toml. Also updated some Actions and docs accordingly. Will hopefully play nice with ReadTheDocs but there could still be issues.


bwpriest:
I understand why the admm wrapper tests are failing - `mpisppy/tests/test_stoch_admmWrapper.py` is trying to import `${REPO_ROOT}/examples/distr/dist_data.py`, but `examples/` is not installed as a package.

I believe this can be fixed by moving `/examples` into `mpisppy/examples`, adding `__init__.py` in appropriate places, and updating the imports, but this is a much larger change that I intended to make. It would also have possibly undesirable outcomes, e.g., `/examples` would then be distributed as `mpisppy.examples` in wheels deployed to pip.

I'm actually not sure why these imports worked in the original `setup.py` version, as `setuptools.find_packages()` does not find the `/examples` or anything therein. So, I am at an impasse as to what to do here. Everything else appears unaffected.

bknueven:
> I understand why the admm wrapper tests are failing - `mpisppy/tests/test_stoch_admmWrapper.py` is trying to import `${REPO_ROOT}/examples/distr/dist_data.py`, but `examples/` is not installed as a package.
> 
> I believe this can be fixed by moving `/examples` into `mpisppy/examples`, adding `__init__.py` in appropriate places, and updating the imports, but this is a much larger change that I intended to make. It would also have possibly undesirable outcomes, e.g., `/examples` would then be distributed as `mpisppy.examples` in wheels deployed to pip.
> 
> I'm actually not sure why these imports worked in the original `setup.py` version, as `setuptools.find_packages()` does not find the `/examples` or anything therein. So, I am at an impasse as to what to do here. Everything else appears unaffected. Any thoughts, @bknueven?

Thanks for your efforts on this @bwpriest. My preference for fixing this test would be to pull / copy the example it needs into the `mpisppy/tests/examples` directory. We already do this, at the risk of be duplicative, but it avoids installing every example as part of the package.

bwpriest:
Thanks @bknueven for your help! I believe this PR is ready. I've verified that the warning discussed in issue #516 is now mitigated.

DLWoodruff (approved):
Thanks! These are valuable changes.

PR #523, opened 2025-05-26
Pylint fun

DLWoodruff:
Fixing up a few things the pylint found in extensions. Most (maybe all) are a matter of style, but worth fixing.


bknueven (commented):
I'd prefer not to forget about some hack-ish things I did in the past...

PR #526, opened 2025-06-11
Xhat fix

bknueven:


mpisppy/cylinders/ph_ob.py
```diff
@@ -89,6 +90,7 @@ def _phboundsolve(self, iternum):
             tee=teeme,
             verbose=verbose,
             warmstart=warmstart,
+            need_solution=False,
         )
```
DLWoodruff, line 94: If there is no solution, what happens to the bounds that solve_one grabs, which it does whether there was a solution or not.  Meanwhile, And Ebound just computes using whatever bounds solve_one has put in there. There is probably something I am missing, but it makes me nervous.

DLWoodruff (commented):
I am nervous about what happens in ph_ob when there is not solution. I left a comment in that file.

DLWoodruff (approved):

.github/workflows/test_pr_and_main.yml
```diff
@@ -196,7 +196,7 @@ jobs:
     - name: Test with nose
       run: |
         nosetests -v mpisppy/tests/test_sc.py
-        mpirun -np 3 -oversubscribe nosetests mpisppy/tests/test_sc.py
+        mpirun -np 3 nosetests mpisppy/tests/test_sc.py
```
bknueven, line 199:
```suggestion
        mpirun -np 3 -oversubscribe nosetests mpisppy/tests/test_sc.py
```

PR #532, opened 2025-06-18
primal_dual_rho: add primal bias option

bknueven:


bknueven (commented):
Add primal-dual rho into rst file

PR #534, opened 2025-06-21
Add return if we have cylinder convergence after iteration 0

bknueven:
No point in continuing `iterk_loop` in this case.


DLWoodruff (approved):
We should have done this years ago.

PR #535, opened 2025-06-26
PHDualSpoke

bknueven:


DLWoodruff (approved):
I like everything about this PR, except the names "DualPH" and "PrimalPH". I prefer "PHDual" and "PHPrimal" for reasons related to the existence of "Dual Simplex"  that I can describe during the zoom meeting, if necessary.

PR #543, opened 2025-07-16
Reimplementation of grad rho

aasgerr:
This PR is a reimplementation of grad rho that consolidates different pieces together with new additions:

- Calculating gradients on the go by caching the objective function gradient expression and evaluating it at the current `x` values or `xhats` as they are available (user specific functionality)

- Allowing scenario dependent denominators for rho calculations in multi-stage stochastic programs.

### DIFF
--- a/mpisppy/extensions/grad_rho.py
+++ b/mpisppy/extensions/grad_rho.py
@@ -23,7 +23,7 @@ class GradRho(mpisppy.extensions.dyn_rho_base.Dyn_Rho_extension_base):
     Gradient-based rho Parameter for Progressive Hedging
     U. Naepels, David L. Woodruff, 2023
 
-    Includes modifications to extend scenario based denominators
+    Includes modifications to extend scenario-based denominators
     to multi-stage problems and calculation of gradients from objective
     expressions on the fly. 
     A. Asger, B. Knueven, 2025


mpisppy/extensions/grad_rho.py
```diff
@@ -0,0 +1,313 @@
...
+        elif self.alpha == 1.0:
+            rhos = {(ndn, i): float(v) for ndn, rho_max in global_rho_maxes.items() for i, v in enumerate(rho_max)}
+        elif self.alpha < 0.5:
+            rhos = {(ndn, i): float(min_v + self.alpha * 2 * (mean_v - min_v))
+                    for ndn in global_rho_mins.keys()
+                    for i, (min_v, mean_v) in enumerate(zip(global_rho_mins[ndn], global_rho_wgted_means[ndn]))}
+        elif self.alpha > 0.5:
+            rhos = {(ndn, i): float(2 * mean_v - max_v + self.alpha * 2 * (max_v - mean_v))
+                    for ndn in global_rho_maxes.keys()
+                    for i, (max_v, mean_v) in enumerate(zip(global_rho_maxes[ndn], global_rho_wgted_means[ndn]))}
```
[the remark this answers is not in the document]
aasgerr: It calculates the alpha weighted in-between values, e.g. if you want higher values than the mean but not the maxes or lower values than the mean but not the mins. This implementation was there in the previous grad_rho too

mpisppy/extensions/grad_rho.py
```diff
@@ -0,0 +1,313 @@
...
+import mpisppy.MPI as MPI
+from mpisppy import global_toc
+import mpisppy.utils.sputils as sputils
+from mpisppy.cylinders.spwindow import Field
+
+class GradRho(mpisppy.extensions.dyn_rho_base.Dyn_Rho_extension_base):
+    """
+    Gradient-based rho from
+    Gradient-based rho Parameter for Progressive Hedging
+    U. Naepels, David L. Woodruff, 2023
```
DLWoodruff, line 24: Add a comment to note the modification authors

mpisppy/extensions/grad_rho.py
```diff
@@ -0,0 +1,313 @@
...
+                        abs(costs[s][ndn, i]/loc_denom[s][ndn, i])
+                        for i, v in enumerate(node.nonant_vardata_list)
+                    ),
+                    dtype="d",
+                    count=nlens[ndn],
+                )
+
+                np.minimum(rho_mins, rhos, out=rho_mins, where=(s._mpisppy_data.prob_coeff[ndn] > 0))
+                np.maximum(rho_maxes, rhos, out=rho_maxes, where=(s._mpisppy_data.prob_coeff[ndn] > 0))
+                rho_wgted_means += s._mpisppy_data.prob_coeff[ndn] * rhos
```
DLWoodruff, line 218: Maybe I am missing something, but if not, the use of wgted is confusing here since the only weight I can see being used is the probability and that is needed to compute the mean.

DLWoodruff (approved):
This is lovely!

PR #554, opened 2025-08-02
Slides

DLWoodruff:
Add a subdirectory to doc for slides to be kept and shared; plus, update agnostic documentation


bknueven (commented):
For causal browsers, it might be nice to include a compiled version of each talk?

bknueven (approved):

PR #555, opened 2025-08-02
PH Relaxed Fixer: do not fix if not converged

bknueven:


DLWoodruff (approved):
This seems OK and maybe I am confused, but it seems like in line 102:
(xb - xvar.lb <= self.bound_tol)
would have almost the same effect, but be one notch clearer if it was:
(xb - relaxed_val <= self.bound_tol)
But now that I type this I can see that "almost" might be doing too much work. Anyway, please merge.

bknueven:
> This seems OK and maybe I am confused, but it seems like in line 102: (xb - xvar.lb <= self.bound_tol) would have almost the same effect, but be one notch clearer if it was: (xb - relaxed_val <= self.bound_tol) But now that I type this I can see that "almost" might be doing too much work. Anyway, please merge.

Interesting, yes. We're trying to test: are xbar, xhat, and relaxed_xhat *all* at a lower bound or upper bound. But maybe it would be better to just test: are xbar, xhat, and relaxed_xhat *all* the same value?

bknueven:
@DLWoodruff I re-worked the fixing logic. Is it clearer now?

PR #556, opened 2025-08-12
Iter0 Warmstarts

bknueven:
#331 fixed a bug when warm-starting PH the initial typical trivial bound would be incorrect if the proximal term was active. To fix this, the warmstart wasn't applied until iteration 1 instead, but this meant the iteration 0 solve was wasted effort.

This PR fixes the same bug but preserves the warmstart being applied at iteration 0 by checking if the proximal term was enabled before reporting the iteration 0 "bound". We also clean up some redundant checking for the `best_bound_obj_val`.


bknueven:
> This was always an issue with reading W, it just seems more stark now that we are doing it pre-iter0: should we check to see if the W is valid? (i.e. E[W] = 0)

It looks like we already check by default: https://github.com/Pyomo/mpi-sppy/blob/a791d8df7c96a283f979742a6dd4fbe2841ae34b/mpisppy/utils/wxbarutils.py#L93

PR #558, opened 2025-08-14
Bounding the scenario dependent denominator in GradRho with a relative bounding factor

aasgerr:
This PR changes the way scenario dependent denominator in GradRho is bounded:

- Earlier it was bounded using `opt.E1_tolerance`, which does not take into account magnitude scale of the non-anticipative variable.
- We now change it to bound with a relative bounding factor using the already existing `grad_rho_relative_bound` and bound it in the form `denom_bound*v.value_`  , where `denom_bound = 1/grad_rho_relative_bound` and `v.value_` is the value of the non-anticipative variable.
- This approach allows to capture the scale of magnitude with a relative bounding value with respect to each non-anticipative variable.

### DIFF
--- a/mpisppy/extensions/grad_rho.py
+++ b/mpisppy/extensions/grad_rho.py
@@ -35,12 +35,8 @@ class GradRho(mpisppy.extensions.dyn_rho_base.Dyn_Rho_extension_base):
         self.opt = opt
         self.alpha = cfg.grad_order_stat
         assert (self.alpha >= 0 and self.alpha <= 1), f"For grad_order_stat 0 is the min, 0.5 the average, 1 the max; {self.alpha=} is invalid."
-        self.multiplier = 1.0
-
-        if (
-            cfg.grad_rho_multiplier
-        ):
-            self.multiplier = cfg.grad_rho_multiplier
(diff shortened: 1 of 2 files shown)


DLWoodruff (approved):

mpisppy/extensions/grad_rho.py
```diff
@@ -66,9 +62,9 @@ def _scen_dep_denom(self, s):
         denom_max = max(scen_dep_denom.values())
 
         for ndn_i, v in s._mpisppy_data.nonant_indices.items():
-            if scen_dep_denom[ndn_i] <= self.opt.E1_tolerance:
-                scen_dep_denom[ndn_i] = max(denom_max, self.opt.E1_tolerance)
-        
+            if (scen_dep_denom[ndn_i]) <= self.denom_bound * v._value:
+                scen_dep_denom[ndn_i] = max(denom_max, self.denom_bound * v._value)
```
bknueven, line 66:
```suggestion
                # denom_max is over all nonants, so if not >0 we will stop
                scen_dep_denom[ndn_i] = max(denom_max, self.denom_bound * v._value)
```

PR #575, opened 2025-10-16
do_decomp is now considered callable

DLWoodruff:
_do_decomp is now do_decomp and it returns its wheel object

### DIFF
--- a/mpisppy/generic_cylinders.py
+++ b/mpisppy/generic_cylinders.py
@@ -132,7 +132,7 @@ def do_decomp(module, cfg, scenario_creator, scenario_creator_kwargs, scenario_d
     """Essentially, the main program for decomposition
 
     Args:
-       module (Python module): the model file with required functions
+       module (Python module or class): the model file with required functions or a class with the required methods.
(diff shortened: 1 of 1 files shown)


mpisppy/generic_cylinders.py
```diff
@@ -128,7 +128,19 @@ def _name_lists(module, cfg, bundle_wrapper=None):
 
 
 #==========
-def _do_decomp(module, cfg, scenario_creator, scenario_creator_kwargs, scenario_denouement, bundle_wrapper=None):
+def do_decomp(module, cfg, scenario_creator, scenario_creator_kwargs, scenario_denouement, bundle_wrapper=None):
+    """Essentially, the main program for decomposition
+
+    Args:
+       module (Python module): the model file with required functions
```
bknueven:
```suggestion
       module (Python module or class): the model file with required functions or a class with the required methods.
```

I think this is correct, and gives the user more flexibility in interfacing with mpi-sppy.

bknueven: Need a description for `cfg`.

PR #608, opened 2026-01-15
Remove duplications of the function `scenario_names_creator`

mosc9575:
The main goal of this PR is to simplify the code base and remove multiple definitions of the functions `scenario_names_creator`.

I tried to replace as many code as possible with the new function and removed quite a few lines of code. I tried to keep all the scenario names the same as before.


DLWoodruff:
There is a lot going on right now and this would touch too many files, so I am going to close this PR.

PR #613, opened 2026-02-14
fixes #612 to keep compatibility with python 3.9

### LINKED ISSUE #612: union of types annotation not supported by python 3.9
after the most recent merge, I am getting the following error on Python 3.9.6
```
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```
at this line:

https://github.com/Pyomo/mpi-sppy/blob/2c1a2d286d3af75857afe37395c06fdbd3493fcf/mpisppy/utils/strings.py#L16

I think this type of annotation is not supported in Python 3.9. The required python should be changed to 3.10 or this annotation changed to something compatible with 3.9.

tvalenciaz:
fixes #612

### DIFF
--- a/mpisppy/utils/strings.py
+++ b/mpisppy/utils/strings.py
@@ -8,12 +8,13 @@
 ###############################################################################
 
 from collections.abc import Iterable
+from typing import Optional
 
 
 def nice_join(
     seq: Iterable,
     separator: str = ", ",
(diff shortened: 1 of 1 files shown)


bknueven (approved):
I'll merge this, although Python 3.9 has reach EOL so I consider support for it in mpi-sppy `main` optional.

Issue #620, opened 2026-02-27
Adding Extensions for ExtensiveForm instances too


tvalenciaz:
It would be nice to have a similar framework to the Extensions that are currently supported for PH-based cylinders for ExtensiveForm instances too. This would allow adding cross-scenario constraints to the model very naturally for both PH and EF instances (among other things the user may be interested in). Pull Request upcoming

PR #621, opened 2026-02-27
Feature/ef extensions

tvalenciaz:
Creates `EFExtensions` modeled on the current `Extension` class for ExtensiveForm objects, which can be of interest as described in #620 . To get a more uniform structure, this PR also modifies the `_do_EF()` logic in `generic_cylinders.py` so it uses the `ExtensiveForm` wrapper instead of creating the ConcreteModel directly.


mpisppy/opt/ef.py
```diff
@@ -164,6 +215,15 @@ def scenarios(self):
         """
         yield from self.local_scenarios.items()
 
+    def _save_active_objectives(self):
+        """ Save the active objectives for use in PH, bundles, and calculation """
+        # maybe move it from spopt to spbase instead?
```
bknueven: Yes, let's do that.

PR #631, opened 2026-03-13
Replace deprecated Pyomo 'undo' with 'reverse' for relax_integer_vars

DLWoodruff:
## Summary
- Pyomo 6.9.3 deprecated the `undo` option in the `core.relax_integer_vars` transformation in favor of `reverse`
- Updated both occurrences: `integer_relax_then_enforce.py` and `nonant_sensitivities.py`

## Test plan
- [x] Run `run_all.py` with the sslp test that uses `--integer-relax-then-enforce` and confirm no deprecation warning


bknueven (approved):
Does this impose a higher minimum Pyomo version than we currently have?

PR #669, opened 2026-04-19
Fix nonant_cost_coeffs for bundles with fixed first-stage vars (#668)

### LINKED ISSUE #668: ProperBundler + --linearize-proximal-terms: KeyError in nonant_cost_coeffs
## Summary

When running a PH cylinder script with proper bundles (`--scenarios-per-bundle=N`) **and** `--linearize-proximal-terms`, the run aborts in `mpisppy.utils.sputils.nonant_cost_coeffs` with:

\`\`\`
File \"mpisppy/phbase.py\", line 732, in attach_PH_to_objective
    scenario._mpisppy_data.nonant_cost_coeffs = sputils.nonant_cost_coeffs(scenario)
File \"mpisppy/utils/sputils.py\", line 1079, in nonant_cost_coeffs
    cost_coefs[nonant_varids[id(var)]] += coef
KeyError: ('ROOT', 3936)
\`\`\`

## Reproducer

From PR #658 (uc_cleanup), with HiGHS:

\`\`\`bash
cd examples/uc
mpiexec -np 2 python -u -m mpi4py cs_uc.py \
    --num-scens=3 --scenarios-per-bundle=3 \
    --max-iterations=1 --default-rho=1 \
    --linearize-proximal-terms \
    --solver-options=\"mip_rel_gap=0.5 threads=1\" \
    --solver-name=appsi_highs
\`\`\`

(With bundling, cs_uc.py auto-disables the cross-scenario-cuts spoke and reduces to plain PH + xhat looper, so the failure is not specific to cs_uc — any PH script that combines proper bundles with the linearized proximal term should hit it.)

## Diagnosis

`nonant_cost_coeffs` walks the active objective and looks up each variable in `_mpisppy_data.nonant_indices`. Inside a bundle EF the active objective has scenario-block variables that aren't in the *bundle's* root-node nonant set (the bundle's nonants are the EF's `ref_vars`, not the per-scenario duplicates), so the lookup misses.

`attach_PH_to_objective` only computes `nonant_cost_coeffs` when `linearize_proximal_terms` is set — without it, the path isn't exercised, which is why the bug hasn't bitten the existing bundling tests (which don't pass `--linearize-proximal-terms`).

## Workaround

Use a quadratic-capable solver (cplex/gurobi/xpress) and drop `--linearize-proximal-terms`. This is fine for the weekly UC regression but blocks running bundles + HiGHS.

## Why this isn't blocking PR #658

PR #658 doesn't actually need this combination to work — `run_uc.py` (weekly) doesn't bundle, and `run_uc_pr.py` (per-PR) uses neither bundles nor cross-scenario cuts. The proper_bundler fix in #658 is now covered by a focused unit test (`mpisppy/tests/test_proper_bundler.py`) that exercises the firstnum/inum translation directly.

DLWoodruff:
## Summary

Fixes #668. When a PH script combined `--linearize-proximal-terms` with `--scenarios-per-bundle` on a model that fixes some first-stage vars (e.g. UC's must-on / must-off generators), `attach_PH_to_objective` called `sputils.nonant_cost_coeffs` and aborted with `KeyError(('ROOT', k))`.

**Root cause**: in the bundle branch, `nonant_cost_coeffs` walked each scenario's `nonant_vardata_list` and assigned `(ndn, i)` keys using the per-scenario position `i` (0..N-1). But the bundle's `_mpisppy_data.nonant_indices` is keyed by 0..M-1 with M < N, because `create_EF` skips fixed-var positions when `proper_bundler` passes `nonant_for_fixed_vars=False`. Lookups for `(ROOT, i)` with `i >= M` then `KeyError`'d against `cost_coefs`.

**Fix**: in the bundle branch, build an explicit per-scenario-position → bundle-position map by enumerating `bundle.ref_vars` (insertion order, skipping `ref_surrogate_vars`) — exactly the ordering `proper_bundler` uses to construct the bundle's `nonantlist`. Then map each per-scenario var's id to the bundle index.

**Test**: focused regression in `mpisppy/tests/test_proper_bundler.py` that builds a 3-scenario bundle from a stub module whose middle nonant is fixed, mocks the SPBase data attrs, and calls `nonant_cost_coeffs` directly. Pre-fix this `KeyError`'d; post-fix it returns the right probability-weighted coefficients.

## Test plan

- [x] `python -m pytest mpisppy/tests/test_proper_bundler.py -v` (6/6 pass)
- [x] `python -m pytest mpisppy/tests/test_sputils.py mpisppy/tests/test_admm_bundler.py` (no regressions)
- [x] End-to-end reproducer:
  ```bash
  cd examples/uc
  mpiexec -np 2 python -u -m mpi4py cs_uc.py \
      --num-scens=3 --scenarios-per-bundle=3 --max-iterations=1 \
      --default-rho=1 --linearize-proximal-terms \
      --solver-options="mip_rel_gap=0.5 threads=1" \
      --solver-name=appsi_highs
  ```
  pre-fix: `KeyError(('ROOT', 3936))`. Post-fix: completes, `BestInnerBound=64140.59 BestOuterBound=64140.49`.


mpisppy/utils/sputils.py
```diff
@@ -1061,13 +1061,30 @@ def nonant_cost_coeffs(s):
...
         for scenario_name in s._ef_scenario_names:
             scenario = s.component(scenario_name)
             for node in scenario._mpisppy_node_list:
                 ndn = node.name
-                for i, v in enumerate(node.nonant_vardata_list):
-                    nonant_varids[id(v)] = (ndn, i)
+                for per_scen_i, v in enumerate(node.nonant_vardata_list):
+                    bundle_key = per_scen_to_bundle.get((ndn, per_scen_i))
+                    if bundle_key is not None:
+                        nonant_varids[id(v)] = bundle_key
```
bknueven, line 1106: These will become the keys into the `cost_coefs` dictionary, but it's not clear that's what we need in the code which consumes the `cost_coefs` dictionary.

DLWoodruff: The consumers should be OK. We added a comment to clarify why we are looping for ndn, even though proper bundles only have ROOT (just in case they someday have more than just ROOT)

bknueven (approved):

Issue #670, opened 2026-04-23
Cut-pool management for cross-scenario and xhat feasibility cuts


DLWoodruff:
## Background

The `CrossScenarioExtension` (`mpisppy/extensions/cross_scen_extension.py`) installs per-iteration cuts into every local scenario's `s._mpisppy_model.benders_cuts` ConstraintList, keyed by `(outer_iter, scen_name)`. Those entries are **never removed** — only the `inner_bound_constr` gets pruned (to keep just the tightest bound). `benders_cuts` accumulates unboundedly over the life of a run.

The proposed xhat feasibility-cut feature (see `doc/xhat_feasibility_cuts_design.md`, addresses issue #601) will install cuts the same way, so it inherits the same unbounded growth.

## Impact

For long runs — especially on problems where each PH iteration triggers a fresh cross-scen cut round, or where an xhatter finds many distinct infeasible candidates — the per-scenario subproblem accumulates constraints indefinitely. Persistent-solver warm starts still work, but solver times grow as the model bloats, and eventually dominate.

## What this issue asks for

A shared cut-pool-management strategy that both extensions can use. Options to consider:

1. **Hard cap per scenario.** Keep at most N cuts; drop the oldest on overflow.
2. **LRU / "not binding for K iterations" prune.** Track whether each cut was at its bound in the last K solves; drop cold ones.
3. **Age-based prune.** Drop cuts older than M outer iterations.

The strategy should live in a small utility (candidate name: `mpisppy.utils.cut_pool`) so both extensions call the same code. Persistent-solver-aware removal (`solver.remove_constraint(...)`) has to be part of it.

## Relation to other work

- `doc/xhat_feasibility_cuts_design.md` (upcoming PR for issue #601) calls out this gap in its "Cut pool management" open question and defers to this issue.
- The `cut_installer` helper proposed in the same design doc is a natural home for the pruning symmetry.

bknueven:
My suggestion would be that we don't use per-scenario cuts at all, and instead use aggregate L-shaped cuts. This also avoids the issue of the number of cuts "blowing up" with the number of scenarios.

Issue #601, opened 2026-01-07
Xhatters should optionally generate feasibility cuts


bknueven:
For problems without complete recourse, an infeasible scenario for xhatters could optionally generate a feasibility cuts to put in every scenario so that solution does not get revisited.

PR #671, opened 2026-04-24
Optional xhatter feasibility cuts, V1 (binary first-stage) — addresses #601

DLWoodruff:
**Design document:** [`doc/xhat_feasibility_cuts_design.md`](./doc/xhat_feasibility_cuts_design.md) — please read that first. It lays out the motivation, the V1/V2/V3 scope table, the code-sharing opportunities with `cross_scen`, and the open decisions (Q1–Q4) the current implementation encodes.

## Summary
First-milestone implementation of issue #601: xhat spokes (`xhatlooper`, `xhatshufflelooper`, `xhatspecific`, `xhatxbar`) may now emit a no-good feasibility cut when their candidate is infeasible in some scenario. The hub installs the cut into every scenario (persistent-solver aware, bundle-safe), so the same xhat is not revisited.

- Off by default. Enabled with `--xhat-feasibility-cuts-count N` (N > 0). N doubles as the per-iteration cap and sizes the shared-memory send buffer.
- **Binary first-stage only.** The hub extension hard-fails at setup time if any nonant is not binary; integer vars with bounds `[0, 1]` are accepted as semantically binary.
- **Two-stage only in V1.** The hub extension hard-fails at setup time if the model is multi-stage: the no-good row is applied positionally against `nonant_indices`, which is only consistent across scenarios when they share the same ROOT nonants (two-stage NAC). Multi-stage needs a branch-aware installer and is deferred to a follow-up milestone (see the design doc).

## What this PR adds
- `Field.XHAT_FEASIBILITY_CUT` on the shared-memory window (row format `[constant, nonant_coefs…]` + trailing count slot).
- `XhatFeasibilityCutExtension` (new hub extension) — binary-only startup check, per-scenario `xhat_feasibility_cuts` ConstraintList, persistent-solver-aware installation.
- Emission path in `XhatBase._try_one`: no-good row with `constant = #{xhat=1} − 1`, `coef_i = 1 − 2·xhat_i`, so `constant + Σ coef_i · x_i ≥ 0` is violated exactly at `xhat`.
- `cfg_vanilla` plumbing (`shared_options` carries the cap; `add_xhat_feasibility_cuts(hub_dict, cfg)` attaches the extension when the flag is positive).
- Wired through the `generic_cylinders` driver (`mpisppy/generic/parsing.py`, `mpisppy/generic/extensions.py`) and `examples/usar/wheel_spinner.py`.
- New user-facing `doc/src/xhat_feasibility_cuts.rst` wired into `index.rst`, plus cross-references from `spokes.rst` and `extensions.rst`.
- New part-1 `examples/run_all.py` entry re-running the USAR wheel spinner with `--xhat-feasibility-cuts-count=3` to guard the full-run plumbing from regressions.

## Scope boundaries
- **V1 (this PR):** binary first-stage, any recourse (MIP OK). Covers UC / scheduling / lot-sizing.
- **V2 (follow-up):** Farkas feasibility cuts via a Pyomo upstream extension to `BendersCutGeneratorData` (currently only emits optimality cuts). Lifts the first-stage-type restriction but requires LP recourse.
- **V3 (not in this design):** integer L-shaped for MIP recourse.

See the scope table in the design doc.

## Follow-ups
- [#670](https://github.com/Pyomo/mpi-sppy/issues/670) — cut-pool management. Today cuts accumulate indefinitely (same behavior as `CrossScenarioExtension` for `benders_cuts`). The per-iter cap keeps growth linear in iterations; #670 will add actual pruning, shared between both features.
- Once this PR **and** [#672](https://github.com/Pyomo/mpi-sppy/pull/672) (`--xhat-from-file`) both merge, a small follow-up can add a `run_all.py` entry that combines them: write a known-infeasible binary xhat to a `.npy`, run `--xhat-from-file <path> --xhat-feasibility-cuts-count=1`, and verify the full emission → install loop end-to-end. That closes the "Known gap" below.

## Test plan
- [x] `ruff check mpisppy/` — clean.
- [x] `python -m unittest mpisppy.tests.test_xhat_feasibility_cuts` — 16 new tests, all pass. Covers startup check (binary passes; `0..1`-bounded integer passes; continuous / integer-not-01 fail with expected error text), cut installation (basic, zero-count no-op, all-zero-row skipped, accumulation), no-good encoding (including the "cut violates at its own xhat" sanity check), and multi-node startup scan.
- [x] `python -m unittest mpisppy.tests.test_ef_ph mpisppy.tests.test_ph_extensions mpisppy.tests.test_proper_bundler` — 38 pre-existing xhat / PH / bundler tests still pass.
- [x] `cd doc && make html` — no new warnings.
- [x] Local USAR smoke: `mpirun -np 3 python -m mpi4py wheel_spinner.py ... --xhat-feasibility-cuts-count=3` completes cleanly; baseline USAR run without the flag is unaffected.
- [x] Full CI.

## Known gap
No end-to-end run with a *crafted infeasible xhat*. The new USAR `run_all.py` entry drives the full WheelSpinner with `--xhat-feasibility-cuts-count` on a binary first-stage model — it guards setup-time plumbing (field registration, binary-only startup check, extension wiring) — but that instance does not actually force an infeasibility, so the emission → install path is covered only by the unit tests in `test_xhat_feasibility_cuts.py`. Happy to add a forced-infeasibility test if reviewers want it.


mpisppy/extensions/xhat_feasibility_cut_extension.py
```diff
@@ -0,0 +1,202 @@
...
+        self._recv_buffer = spcomm.register_recv_field(
+            Field.XHAT_FEASIBILITY_CUT, ranks[0]
+        )
+
+    def sync_with_spokes(self):
+        if self._recv_buffer is None:
+            return
+        if self._recv_buffer.is_new():
+            self._install_cuts(self._recv_buffer.array())
+
```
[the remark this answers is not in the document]
DLWoodruff: Good catch — this is a real crash. Every xhat inner-bound spoke advertises `Field.XHAT_FEASIBILITY_CUT` via `XhatInnerBoundBase.send_fields`, and the generic driver appends both `xhatshuffle_spoke` and `xhatxbar_spoke`, so `--xhatshuffle --xhatxbar --xhat-feasibility-cuts-count N` would trip the assert.

Fixed in 716b7a1e: `register_receive_fields` now registers one recv buffer per emitting rank (`self._recv_buffers`), and `sync_with_spokes` installs cuts from whichever buffer is new. A no-good cut from any xhatter is valid for every scenario, so installing from all emitters is correct. Added tests for multiple emitters registering and installing from each buffer.

mpisppy/extensions/xhat_feasibility_cut_extension.py
```diff
@@ -0,0 +1,202 @@
...
+        so the same row applied positionally to each scenario lands
+        coefficients on unrelated variables. A multi-stage-correct
+        installer needs to group cuts by branch; that is deferred to a
+        follow-up milestone.
+        """
+        if getattr(self.opt, "multistage", False):
+            raise RuntimeError(
+                "--xhat-feasibility-cuts-count > 0 is two-stage only in "
+                "V1. Multi-stage support is planned as a follow-up "
+                "milestone (the install side needs to group cuts by "
```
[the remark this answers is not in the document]
DLWoodruff: You're right — the code and design doc are the source of truth, and V1 is two-stage only (`_assert_two_stage` hard-fails on multi-stage). The PR description bullet was stale; I've corrected it to state two-stage only in V1, with the positional-`nonant_indices` rationale and multi-stage deferred to a follow-up milestone. No code change — the multi-stage guard already matches the design.

mpisppy/cylinders/xhatbase.py
```diff
@@ -108,7 +116,32 @@ def _try_file_xhat(self):
         if self.cylinder_rank == 0:
             print(f"[xhat-from-file] evaluating {path!r} "
                   f"({n_nonants} nonants)")
-        Eobj = self.opt.evaluate(nonant_cache)
+        try:
+            Eobj = self.opt.evaluate(nonant_cache)
+        except Exception:
+            Eobj = None
```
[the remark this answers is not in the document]
DLWoodruff: Agreed. Fixed in 716b7a1e: the `except` around `self.opt.evaluate(nonant_cache)` now prints the exception on rank 0 before degrading to `Eobj=None`, so a real evaluate/solver failure is surfaced instead of hiding behind a bare `Eobj=None`. Kept the graceful-degradation behavior (run continues) but the failure is now visible.

doc/designs/xhat_feasibility_cuts_design.md
```diff
@@ -0,0 +1,492 @@
...
+`pyomo.contrib.benders.benders_cuts.BendersCutGeneratorData.generate_cut()`
+**hard-errors** when a subproblem is not optimal:
+
+```python
+if res.solver.termination_condition != pyo.TerminationCondition.optimal:
+    raise RuntimeError('Unable to generate cut because subproblem failed
+                        to converge.')
+```
+
+That is, it generates **optimality cuts only** from optimally-solved
```
bknueven: Claude is wrong about this, but I understand why it's confused. Pyomo's Benders machinery generated blended optimality and feasibility cuts in a single pass.

doc/src/xhat_feasibility_cuts.rst
```diff
@@ -0,0 +1,163 @@
...
+  node. The current installer applies one cut row uniformly to every
+  scenario, which is only valid in two-stage; that's why V1 hard-fails
+  at setup on a multi-stage model. Tracking as a follow-up alongside
+  V2.
+- Lifting the binary-only restriction requires generating **Farkas
+  feasibility cuts** from the dual ray of an infeasible second-stage
+  LP. The upstream
+  ``pyomo.contrib.benders.benders_cuts.BendersCutGeneratorData``
+  currently only produces optimality cuts; supporting the
+  infeasibility case is a Pyomo PR. Once that lands, the xhatter will
```
bknueven: But, that said, it should probably be revisited, or we implement our own mechanism.

bknueven (approved):
No-good cuts cost nothing, so this is a low-risk feature that could help certain classes of SMIPs. I say we ship it.

doc/designs/xhat_feasibility_cuts_design.md
DLWoodruff: You're right — thanks. I traced it through `_setup_subproblem`: it relaxes every subproblem constraint with a nonnegative violation variable `_z` (the GLM99 formulation the module cites) and minimizes it, so the subproblem is always feasible and `generate_cut()` returns a single *blended* optimality/feasibility cut — `_z = 0` gives an optimality cut, `_z > 0` certifies and cuts off recourse infeasibility. The `RuntimeError` is a genuine-solver-failure guard, not an optimality-only restriction.

Corrected the design doc's 'pyomo Benders gap' section: it now says pyomo already yields a valid cut for the LP-recourse infeasibility case (no Pyomo PR needed), with the two real caveats — dual-based so LP recourse only, and L1-penalty style rather than a dedicated Farkas cut, so worth revisiting. Pushing shortly.

doc/src/xhat_feasibility_cuts.rst
DLWoodruff: Agreed — updated the rst bullet to match. It now says pyomo already emits *blended* optimality/feasibility cuts for LP recourse in a single pass (so no Pyomo PR is needed for the infeasibility case), and flags that the L1-penalty blended cut is worth revisiting — we may tighten it or generate our own dedicated Farkas feasibility cut from the infeasible subproblem's dual ray. V1's binary-only no-good cut is unchanged; this only concerns the LP-recourse V2 path.

DLWoodruff: Tracked the V2 mechanism decision (reuse pyomo's blended cut / tighten it / roll our own Farkas cut from the dual ray) as #796 so it isn't buried in doc prose.

doc/designs/xhat_feasibility_cuts_design.md
DLWoodruff: Full disclosure: Claude wrote that reply.

doc/src/xhat_feasibility_cuts.rst
DLWoodruff: Full disclosure: Claude wrote that reply.

PR #679, opened 2026-05-07
rho setters: aggregate per-sub-scenario nonants in proper bundles (closes #673)

### LINKED ISSUE #673: Grad rho and sep rho and rc rho with bundles
(Maybe we should deprecate rc rho and sep rho. RC rho has not shown to be effective, and sep rho should be subsumed into grad rho.)

The objective coefficient / gradient for bundles will be just have the representative nonant, whereas the objective coefficient / gradient should be across *all* nonants in the bundle.

DLWoodruff:
## Summary

Closes #673. Fixes all three rho setters that the issue called out as using
the bundle's representative ref Var instead of aggregating across every
per-sub-scenario nonant in a consensus group, plus deprecation notices for
the two rho setters bknueven flagged as candidates for removal.

`sputils.create_EF` picks the *first* sub-scenario's nonant Var as the
bundle ref and ties the rest by NA equality constraints. The bundle
objective references each sub-scenario's *own* nonant Vars; the ref Var
typically does not appear in the objective at all (just in the NA
constraints). Probing the ref alone therefore captured one
representative sub-scenario's contribution plus the NA multiplier —
not the bundled rate of change.

This PR replaces ref-only probing with **consensus-group aggregation**
in `grad_rho`, `sensi_rho`, and `reduced_costs_rho`. The same helper
(`_bundle_consensus_groups`) that maps a bundle nonant position
`(ndn, k)` to the list of per-sub-scenario nonant Vars feeds all three
fixes; it parallels the bundle branch already in
`sputils.nonant_cost_coeffs`.

### grad_rho

- `_get_grad_exprs`: differentiate w.r.t. every per-sub-scenario nonant
  Var, sum the partial expressions per bundle position.
- `_eval_grad_exprs`: propagate xhat to all per-sub-scenario nonants at
  each position (NA-feasible by construction).

### sensi_rho (`mpisppy/utils/nonant_sensitivities.py`)

- New private helper `_bundle_consensus_groups`.
- KKT probe: build `y_vec` as a 0/1 indicator over every per-sub-scenario
  nonant Var index — a feasible consensus perturbation. The Rayleigh-
  quotient form `(g'K^{-1}y / y'K^{-1}y)` is scale-invariant, so the
  result is the rate of bundle-objective change per unit consensus shift.

### reduced_costs_rho (`mpisppy/cylinders/reduced_costs_spoke.py`)

- New `_consensus_rc_sum` sums `scenario.rc[v]` over the consensus group
  at each bundle position (NA multipliers cancel in the sum). Falls back
  to `scenario.rc[ref_var]` for unbundled scenarios, preserving existing
  behavior bit-for-bit.
- `vars_to_load` for persistent solvers now covers every per-sub-scenario
  nonant in the bundle case so all `rc` entries are loaded.
- New `_assert_consensus_rc_loaded` runtime guard verifies every Var in
  every consensus group has an `rc` entry before the aggregation reads
  it. Catches a future regression where `vars_to_load` is restricted
  back to ref Vars (or any other path that drops sub-scenario nonants).

### Bundler stashes the consensus_groups mapping

Per bknueven's review feedback that this mapping is something the
bundler should provide rather than each rho setter rederiving:
`sputils.create_EF` (both the multi-scenario and singleton branches)
now stashes `(ndn, k) -> [per-sub-scenario Var, ...]` on the EF as
`consensus_groups` at construction time, as a side effect of the
existing per-scenario / per-node / per-i loop.
`_bundle_consensus_groups` collapses to an O(1) lookup for bundles
and synthesizes singleton groups for unbundled scenarios, so callers
no longer branch on `hasattr(s, "_ef_scenario_names")` just to
construct the mapping:

- `grad_rho._get_grad_exprs` / `_eval_grad_exprs` lose their
  bundle-vs-unbundled forks; the unbundled case is the singleton-group
  degeneration of the bundle aggregation loop.
- `reduced_costs_spoke` drops the per-iteration
  `consensus_groups_by_sub` cache (the helper is now cheap) and
  flattens `vars_to_load` uniformly.
- `nonant_sensitivies` drops the `consensus_groups is None` branch on
  the KKT probe.

The `_consensus_rc_sum` / `_assert_consensus_rc_loaded` helpers stay
tolerant of `None` so the existing focused tests against those entry
points continue to exercise both paths.

### SepRho / ReducedCostsRho deprecation

Per bknueven's note in #673: `SepRho` is expected to be subsumed into
`GradRho`, and reduced-cost rho has not shown to be effective in
practice. Both classes now emit a `DeprecationWarning` at instantiation
so downstream projects have lead time to migrate. Removal scheduled for
a future PR.

## Tests

Four new solver-free unit-test files. Each was confirmed to fail under a
simulated representative-only behavior (stash-revert or mocked helper).

- `mpisppy/tests/test_grad_rho_bundles.py` — 5 tests
  - linear-obj baseline; bundled equivalence with explicit
    "not representative-only" guard; quadratic-obj eval at xhat with
    primed stale per-sub-scenario var values.
  - `TestBundledRhoMatchesUnbundledMean` — end-to-end
    `_compute_and_update_rho` on the same 3 scenarios unbundled and as
    one bundle (`MPI.COMM_SELF` for the per-node Allreduce; fake
    xbars/rho/prob_coeff on the SPBase side). All-positive linear
    coefficients, `alpha=0.5`, `EF_prob=1`, deterministic shared denom:
    bundle rho exactly equals the conditional-prob-weighted mean of
    unbundled rhos, with a guard that it is *not* equal to scen0's
    value alone (representative-only regression).
  - `TestBundleAndUnbundledDivergeUnderSignCancellation` — same
    harness with `c = [[+5,+2],[-5,+2],[+1,+2]]`. Position 1 (no sign
    disagreement) still exact-matches; position 0 sees the bundle rho
    shrink to ~0.33 vs the unbundled mean ~3.67 (an 11x divergence)
    because the bundle averages gradients *before* `|.|` while the
    unbundled mean averages `|.|`s. Pins both numbers and the
    triangle-inequality bound. Documents the real algebraic property
    so future readers don't expect the two paths to give similar rho
    values in general.
- `mpisppy/tests/test_sensi_rho_bundles.py` — 2 unit tests + 2
  `skipUnless(ipopt_available)` end-to-end smoke tests.
  - Helper returns N Vars per position for an N-sub-scenario bundle, ref
    is one of them, non-ref sub Vars present.
  - Smoke: pipeline completes; one-sub-scenario bundle agrees with the
    standalone scenario.
- `mpisppy/tests/test_reduced_costs_rho_bundles.py` — 6 tests
  - `_consensus_rc_sum`: unbundled returns ref Var rc unchanged; bundle
    returns the consensus sum across the group, with explicit
    "not equal to ref-only" guard.
  - `_assert_consensus_rc_loaded`: unbundled is a no-op even with an
    empty rc Suffix; passes when every consensus Var has rc; fails when
    only ref Vars have rc; fails when one sub-scenario Var is missing.
- `mpisppy/tests/test_rho_deprecations.py` — 2 tests
  - `SepRho` and `ReducedCostsRho` each emit a `DeprecationWarning`
    naming the class and the issue reference.

All four new test files wired into `run_coverage.bash` (serial phase)
and the unit-tests pytest invocation in
`.github/workflows/test_pr_and_main.yml` alongside `test_jensens.py` /
`test_proper_bundler.py`.

## Test plan

- [x] `ruff check .` clean
- [x] `python -m pytest mpisppy/tests/test_grad_rho_bundles.py -v` — 5 pass
- [x] `python -m pytest mpisppy/tests/test_sensi_rho_bundles.py -v` —
      2 pass + 2 skipped without Ipopt; 4 pass with Ipopt
- [x] `python -m pytest mpisppy/tests/test_reduced_costs_rho_bundles.py -v` — 6 pass
- [x] `python -m pytest mpisppy/tests/test_rho_deprecations.py -v` — 2 pass
- [x] `python -m pytest mpisppy/tests/test_gradient_rho.py -v` — existing smoke still green
- [x] `python -m pytest mpisppy/tests/test_proper_bundler.py mpisppy/tests/test_admm_bundler.py mpisppy/tests/test_jensens.py mpisppy/tests/test_pickle_bundle.py` — 68 pass (covers the bundler / consensus-groups refactor end of things)
- [x] Coverage CI green on all four new files


mpisppy/cylinders/reduced_costs_spoke.py
```diff
@@ -160,17 +227,45 @@ def extract_and_store_reduced_costs(self):
         # would probably need additional info about where scenarios disagree
         rc = np.zeros(self.nonant_length)
 
+        # _bundle_consensus_groups walks every sub-scenario/node/var in the
+        # bundle; cache once per `sub` and reuse below in the second pass that
+        # populates _scenario_rc_buffer.
+        consensus_groups_by_sub = {
```
bknueven: Is this something the bundle-er should be providing for us?

mpisppy/extensions/grad_rho.py
```diff
@@ -129,32 +130,97 @@ def _scen_indep_denom(self):
...
-                                mode=Modes.reverse_symbolic,
-                                )
+            active_obj = sputils.find_active_objective(s)
+
+            if hasattr(s, "_ef_scenario_names"):
+                # Shared helper builds the bundle position -> per-sub-scenario
+                # Vars mapping (see mpisppy/utils/nonant_sensitivities.py); we
+                # then flatten it into parallel wrt_vars/wrt_keys so we can
+                # differentiate once and re-aggregate the partials by key.
+                per_scen_vars = _bundle_consensus_groups(s)
```
bknueven: Should the bundle-er just provide / store this?

bknueven (commented):
Should the data we need just go in `s._mpisppy_data`, and then therefore replicated on single scenarios so we don't need to branch anywhere?

mpisppy/cylinders/reduced_costs_spoke.py
DLWoodruff: Done in 0ae8bcd. `sputils.create_EF` now stashes the densified `(ndn, k) -> [per-sub-scenario Var, ...]` mapping on the EF as `consensus_groups` at construction time. `_bundle_consensus_groups` collapses to an O(1) lookup for bundles and synthesizes singleton groups for unbundled scenarios, so callers no longer branch on `hasattr(s, "_ef_scenario_names")` just to construct the mapping. The `consensus_groups_by_sub` cache here goes away — the helper is cheap enough now to call inline in both passes — and `vars_to_load` is uniformly a flatten of `consensus_groups.values()`.

mpisppy/extensions/grad_rho.py
DLWoodruff: Done in 0ae8bcd — `sputils.create_EF` (both the multi-scenario and singleton branches) now stashes the densified `consensus_groups` mapping on the EF as a side effect of the existing per-scenario / per-node / per-i loop, so `_bundle_consensus_groups` becomes an O(1) attribute read. The construction here also collapses to a single uniform path (the unbundled case is the singleton-group degeneration of the bundle aggregation loop) — no more `if hasattr(s, "_ef_scenario_names"):` branch, and `_bundle_per_scen_vars` is just `_per_scen_vars` now since it's populated for every scenario.

PR #686, opened 2026-05-13
stoch-admm phase A: first_stage_cost / first_stage_varlist module hooks

DLWoodruff:
## Summary

Phase A of the ADMM user-API automation plan. The phased plan A–E itself is folded into this PR as `doc/designs/admm_user_api_automation_design.md` (see the `design:` commit at the end of the branch).

Eliminate the asymmetry where `--admm` users must NOT call `sputils.attach_root_node` in `scenario_creator` but `--stoch-admm` users MUST. When the model module defines `first_stage_cost(scenario)` and `first_stage_varlist(scenario)` callables, `Stoch_AdmmWrapper` (and `AdmmBundler`, for the bundled path) call `attach_root_node` themselves; `scenario_creator` just returns the constructed model. The legacy path is preserved for backward compatibility.

**Strict both-or-neither contract.** Per-scenario error matrix, checked right after `scenario_creator` returns:

|                              | both hooks   | neither hook |
|------------------------------|--------------|--------------|
| `attach_root_node` called    | `RuntimeError` | OK (legacy)    |
| not called                   | OK (hooks)     | `RuntimeError` |

The "hooks + manual call → error" choice is deliberately strict (rather than silently overwriting the user's call): users learn the contract instead of half-migrating undetected. The "neither hook + no call" error replaces a cryptic deep assertion in `assign_variable_probs` with a message pointing at both remediation options.

### Files

- **`doc/designs/admm_user_api_automation_design.md`**: phased plan A–E with rationale, per-phase scope, back-compat strategy, and resolved open questions.
- **`mpisppy/utils/stoch_admmWrapper.py`**: hook parameters + per-scenario error matrix.
- **`mpisppy/utils/admm_bundler.py`**: same plumbing for the bundled path (constructs scenarios in its own `scenario_creator` method; runs the matrix per scenario inside the bundle).
- **`mpisppy/generic/admm.py`**: `setup_stoch_admm` and `setup_stoch_admm_with_bundles` discover hooks on the module and enforce both-or-neither.
- **`examples/stoch_distr/stoch_distr.py`**: migrate to the new API. `scenario_creator` stashes the first-stage varlist on the model; new hook functions read it back. `consensus_vars_creator` is refactored to read first-stage vars via `first_stage_varlist()` instead of walking `model._mpisppy_node_list` (which no longer exists at that point — the wrapper hasn't attached the root yet).
- **`examples/stoch_distr/stoch_distr_admm_cylinders.py`**: the custom driver forwards the module's hooks so `stoch_distr_ef.py` (which uses this driver) continues to work after the migration.
- **`doc/src/generic_admm.rst`**: rewrites "Extending to Stochastic ADMM" around the hook-based API as the recommended path; preserves the legacy manual-attach pattern as a fallback subsection.
- **`mpisppy/tests/test_stoch_admmWrapper.py`**: 6 new tests covering both paths, all four error-matrix cells, and the `setup_stoch_admm`-level half-hooks error.

### Test plan

- [x] \`python -m pytest mpisppy/tests/test_stoch_admmWrapper.py --deselect …test_values\` — 16/16 (incl. 6 new) pass (the deselected \`test_values\` is pre-existing flaky under subprocess capture, unrelated; fixed in #688)
- [x] \`python -m pytest mpisppy/tests/test_admm_bundler.py\` — 26/26 pass (legacy \`AdmmBundler\` path covered via the tests-examples copy of \`stoch_distr.py\`, which keeps the manual \`attach_root_node\` call)
- [x] \`examples/stoch_distr/stoch_admm_stage2ef.bash\` (new hooks path via \`generic_cylinders --stoch-admm\`) — converges
- [x] \`examples/stoch_distr/go.bash\` (legacy custom-driver path, now forwarding hooks) — converges
- [x] \`cd doc && make html\` — clean
- [x] \`ruff check\` — clean

### Follow-up

- Supersedes the doc clarification from commit \`24bf5802\` in PR #685 ("clarify attach_root_node contract differs for --admm vs --stoch-admm").
- Next phases (B–E) are in the design doc and will be PRs in sequence after this one merges.


examples/stoch_distr/stoch_distr.py
```diff
@@ -249,16 +269,18 @@ def consensus_vars_creator(admm_subproblem_names, stoch_scenario_name, inter_reg
             print(f"WARNING: {admm_subproblem_name} has no consensus_vars")
             consensus_vars[admm_subproblem_name] = list()
     
-    # now add the parents. It doesn't depend on the stochastic scenario so we chose one and
-    # then we go through the models (created by scenario creator) for all the admm_stoch_subproblem_scenario 
-    # which have this scenario as an ancestor (parent) in the tree
+    # Now add the original problem's first-stage variables (factory
```
emglista, line 272: I know there was discussion in the past about having to define this consensus_vars dict with the first-stage variables (the model has to be instantianted an extra time here (line 280)).  It would be cleaner to do this combination of first-stage and admm vars in the wrapper, if that is an easily implementable option? (Backwards compatibility may be an issue?)  I see that making the consensus_vars list more user-friendly is in a future step of the design, so maybe this is already in the works.

mpisppy/utils/stoch_admmWrapper.py
```diff
@@ -92,27 +92,65 @@ def __init__(self,
...
+            if has_first_stage_hooks:
+                if already_attached:
+                    raise RuntimeError(
+                        f"scenario {sname!r}: first_stage_cost and "
+                        f"first_stage_varlist hooks are defined, so "
+                        f"scenario_creator must NOT call "
+                        f"sputils.attach_root_node.  Remove the "
+                        f"attach_root_node call from scenario_creator."
+                    )
+                sputils.attach_root_node(
```
emglista, line 142: What if we want to attach the surrogate_nonant_list-- should users use the old format where you call attach_root_node in the scenario_creator?

examples/stoch_distr/stoch_distr.py
DLWoodruff: Good catch — this is something Phase B should fix, and you're right that Phase B as originally scoped didn't quite cover it. The throwaway scenario instantiation only exists because `consensus_vars_creator` runs *before* the wrapper attaches the root, so `_mpisppy_node_list` isn't available to walk. Now that Phase A makes `first_stage_varlist(scenario)` available to the wrapper at construction time, the wrapper can do the first-stage + admm union itself and `consensus_vars_creator` can return only the admm-consensus vars.

I've folded this into the Phase B section of the design doc as item **B.2** (pushed in 27a4ace4 on this branch). Back-compat is preserved: the legacy (no-hooks) path keeps the user-side merge.

mpisppy/utils/stoch_admmWrapper.py
DLWoodruff: Real gap — thanks for flagging it. Phase A's `attach_root_node` call is positional and only passes `firstobj` + `varlist`, so users who need `surrogate_nonant_list` (or `nonant_ef_suppl_list`) are forced back to the legacy `attach_root_node`-in-`scenario_creator` form. Rather than leave it that way, I've added item **B.3** to Phase B: optional `first_stage_surrogate_nonant_list` / `first_stage_nonant_ef_suppl_list` module hooks that the wrapper forwards to `attach_root_node`. Validation: defining either advanced hook without the two core Phase A hooks is a hard error.

Pushed in 27a4ace4. For Phase A I'll add a doc note that surrogates currently require the legacy path until B.3 lands.

PR #709, opened 2026-05-17
incumbent spokes: optionally write solution on each new best (#285)

### LINKED ISSUE #285: Incumbent finders should optionally write a new solution immediately

DLWoodruff:
## Summary

- Fixes #285 (good-first-issue). The first-stage solution was previously written only at end-of-run via `WheelSpinner.write_first_stage_solution`. This PR adds an opt-in CLI flag so xhat-finder spokes snapshot the first-stage solution on every new best inner bound — useful for long runs where downstream tooling wants the current incumbent without waiting for the cylinder system to finish.
- New flag: `--incumbent-on-improvement-filename-prefix <prefix>` (registered in `Config.popular_args`). Default `None` (disabled).
- When set, `_BoundSpoke.update_if_improving` writes `<prefix>_<NNNN>.csv` and `<prefix>_<NNNN>.npy` on each new best, where `<NNNN>` is a per-spoke zero-padded counter starting at `0000` so successive incumbents don't overwrite each other.
- Forwarded via `cfg_vanilla.shared_options` so all hub/spoke setups built through the vanilla factories pick it up automatically.
- Fail-soft: if `write_first_stage_solution` raises (typically because the spoke uses `update_best_solution_cache=False` and the cache isn't populated — `xhatlooper`, `xhatshuffle`), warn once on rank 0 and disable further per-improvement writes from that spoke. Happy paths: `xhatxbar` and `xhatspecific` (which use `update_best_solution_cache=True`, so `best_solution_cache` is populated before the writer is called).

## Open questions

- Counter-suffixed filenames vs. rolling overwrite (`<prefix>.csv` overwritten on each improvement): I went with counters so the trajectory is recoverable. Easy to change if rolling is preferred.
- Could also write the full tree solution via `write_tree_solution` instead of just first-stage; held off because tree writes are heavier and most users want the first-stage candidate.

## Test plan

- [x] `ruff check` clean on changed files
- [x] New unit tests in `mpisppy/tests/test_incumbent_writing.py` (13 tests, all passing locally) covering:
  - `Config.popular_args()` registers `incumbent_on_improvement_filename_prefix` with default `None`
  - `cfg_vanilla.shared_options` forwards the prefix into the options dict on both hub and spoke
  - `InnerBoundSpoke._maybe_write_incumbent_on_improvement` no-ops when prefix is `None` or the per-spoke disabled flag is set
  - Happy path writes `<prefix>_<NNNN>.csv` then `<prefix>_<NNNN>.npy` (with the npy serializer), increments the counter, and zero-pads to four digits; counter advances across repeated calls
  - Fail-soft branch: a `RuntimeError` from `write_first_stage_solution` produces exactly one warning on rank 0, sets the disabled flag, does NOT bump the counter, and short-circuits subsequent calls; non-zero ranks stay silent but still set the disabled flag
- [x] Test file wired into both `run_coverage.bash` and `.github/workflows/test_pr_and_main.yml` unit-tests job so codecov/patch picks up the new lines
- [ ] CI on this PR

### DIFF
--- a/.github/workflows/test_pr_and_main.yml
+++ b/.github/workflows/test_pr_and_main.yml
@@ -1063,7 +1063,8 @@ jobs:
             mpisppy/tests/test_solver_spec.py \
             mpisppy/tests/test_extensions.py \
             mpisppy/tests/test_jensens.py \
-            mpisppy/tests/test_proper_bundler.py
+            mpisppy/tests/test_proper_bundler.py \
+            mpisppy/tests/test_incumbent_writing.py
 
       - name: Upload coverage data
         if: always()
--- a/mpisppy/cylinders/spoke.py
+++ b/mpisppy/cylinders/spoke.py
@@ -11,9 +11,11 @@ import abc
 import time
 import os
 import math
+import warnings
 
 from mpisppy.cylinders.spcommunicator import SPCommunicator, SendCircularBuffer
 from mpisppy.cylinders.spwindow import Field
+from mpisppy.utils import sputils
 
 
 class Spoke(SPCommunicator):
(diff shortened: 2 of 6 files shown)


bknueven:
Is there an issue with `xhatshuffle` which this PR brings to light? I'm a bit confused that all paths aren't "happy".

Issue #726, opened 2026-05-24
Flexible ranks: replace fullcomm.allgather layout exchange with a scalable scheme before production use


DLWoodruff:
## Context

Flexible (unequal) per-cylinder rank assignments — see
`doc/designs/flexible_rank_assignments.md`.

The unequal-rank communication path exchanges each rank's buffer layout via a
single **`fullcomm.allgather`** (in practice, the existing `SPWindow` layout
exchange run on `fullcomm` instead of `strata_comm`). This is correct and
effectively zero new code, and it is fine at development/test scale.

## Problem

It is an O(N) **startup** allgather with O(N)-per-rank layout storage, where N
is the total rank count. Total rank counts in the **thousands** are a real
operating regime for mpi-sppy, so this exchange must not ship to production as
is.

## What to do

Replace `fullcomm.allgather` with the scalable scheme the design doc prefers,
before flexible ranks is documented or recommended for production use:

- **Two-level exchange**: `cylinder_comm.allgather` within each cylinder,
  followed by a small cross-cylinder gather over one anchor rank per cylinder;
  or
- **Local compute**: reconstruct each peer rank's layout from the static
  per-cylinder rank counts, the scenario-to-rank map, and the field
  registration order (no communication), which is clean for the uniform
  two-stage case and may need a small exchange for non-uniform/multistage.

This is a **localized** change: it only affects how `strata_buffer_layouts` is
populated at startup, not the multi-source reader, so it can be reviewed and
scale-tested on its own.

## Notes

- This is a **prerequisite before the feature is recommended for production use**, not a feature phase — it does not block the
  remaining flexible-rank phases (coherence policy, multistage/FWPH, APH).
- References: the "Option D" layout-exchange note and the "Gate reliance on the
  feature with an MPI CI matrix" section in the design doc.

DLWoodruff:
Closing as won't-fix: on re-examination, the scaling analysis does not support replacing the flat `fullcomm.allgather` layout exchange.

- It runs **once**, at startup, inside window creation — never on the RMA hot path.
- An allgather's latency is O(log N) rounds in any production MPI implementation; the O(N) part is per-rank *data received*, which is unavoidable because the result itself is O(N): every rank legitimately needs the layout of any peer rank its overlap maps touch.
[...]

PR #731, opened 2026-05-28
flex-ranks phase 2: communication layer (unequal-rank fullcomm window + multi-source reader)

DLWoodruff:
> **📚 Stacked PRs — flexible rank assignments**
> Reviewed and merged bottom-up; each PR targets the one below it.
>
> 1. **Pyomo/mpi-sppy#725 — Phase 1: infrastructure** — ✅ merged
> 2. **Phase 2: communication layer** ← _this PR_ (targets `main`)
> 3. **DLWoodruff/mpi-sppy-1#13 — Phase 3a: per-field coherence (DUALS strict)** (targets `flex-ranks-phase2-communication`)
> 4. **DLWoodruff/mpi-sppy-1#14 — Phase 3b: CLI rank-ratio flags** (targets `flex-ranks-phase3-coherence`)
> 5. **DLWoodruff/mpi-sppy-1#15 — Phase 4a: two-stage xhat fields** (targets `flex-ranks-phase3b-cli`)
>
> 6. **DLWoodruff/mpi-sppy-1#16 — Phase 4b: multistage per-node xhat sourcing** (targets `flex-ranks-phase4a-xhat-2stage`)
> 7. **DLWoodruff/mpi-sppy-1#17 — Phase 5: APH (asynchronous sender) verification** (targets `flex-ranks-phase4b-multistage`)
>
> _Later phases will be appended here as they open._

---

Second phase of the flexible-rank design (`doc/designs/flexible_rank_assignments.md`).
Adds the communication layer that actually runs cylinders at unequal rank
counts. **Gated-additive: the equal-rank path is unchanged** — all of this is
reached only when a cylinder's `rank_ratio` differs from 1.0.

## What's here

- **`spin_the_wheel.py`** — on a non-uniform rank-ratio request, apportion the
  rank pool into contiguous global-rank blocks, build `cylinder_comm` only
  (no `strata_comm`, which is undefined when cylinders differ in size), and
  pass `strata_comm=None`. The phase-1 `NotImplementedError` gate is removed.
- **`rank_apportionment.py`** — `cylinder_bases` / `rank_to_cylinder` for the
  contiguous-block layout.
- **`SPCommunicator`** — when `strata_comm is None`: window on `fullcomm`,
  peers addressed by global rank. `get_receive_buffer` gains a flex dispatch —
  single-source (from a peer cylinder's base rank) for global/scalar fields,
  multi-source overlap-map assembly for per-scenario fields.
- **Layout exchange** is `fullcomm.allgather` (interim; the scalable
  two-level/local-compute replacement is a pre-advertisement release gate,
  tracked in Pyomo/mpi-sppy#726).

## Coherence — two RMA bugs found while testing (this is the flaky corner the design flags)

- **Lockstep:** consumers like `xhatshufflelooper` require every reader rank to
  agree on "is_new" so the collectives they run on new data stay matched on
  `cylinder_comm`. The multi-source reader mirrors the equal-rank cross-rank
  `write_id` agreement rather than accepting each source independently.
- **Atomicity:** each source's whole field is read in **one** `Get` (data +
  trailing `write_id` from a single snapshot), committed into the buffer only
  after the agreement check passes. Reading data and id in separate `Get`s
  raced the writer (pre-write NaN data paired with a fresh id, accepted as new).

## Tests (np=6, both on `fullcomm` windows; `-oversubscribe`)

- **`test_spwindow_multisource.py`** — builds an `SPWindow` directly on
  `COMM_WORLD` and asserts exact multi-source assembly from several remote
  ranks (incl. a reader rank straddling a writer boundary). Isolates the RMA
  primitive from the hub/spoke stack.
- **`test_flexible_rank_cylinders.py`** — farmer PH hub + xhatshuffle spoke at
  **4+2 and 2+4**. The only local-sized field crossing cylinders is
  `NONANTS_VALS`; both asymmetry directions must agree and reach the
  extensive-form optimum.

Both wired into `run_coverage.bash` and the `test-cylinders` CI job. Equal-rank
regression (`test_with_cylinders` np=2) unchanged.

### DIFF
--- a/mpisppy/cylinders/spcommunicator.py
+++ b/mpisppy/cylinders/spcommunicator.py
@@ -536,6 +536,42 @@ class SPCommunicator:
         self.window.put(buf.window_array(), field)
         return
 
+    def _write_ids_agree(self, new_id: int, synchronize: bool) -> bool:
+        """Cross-rank write_id agreement on ``cylinder_comm`` (shared by every
+        reader: the equal-rank path and both unequal-rank helpers).
+
+        Consumers that run collectives on freshly-received data (Eobjective
+        Allreduce, ROOT bcast, ...) must enter them in lockstep, so every reader
+        rank has to agree on whether a read is "new". A synchronous writer
+        stamps all of its ranks at one write_id, so when ids agree every rank
+        computes the same answer; a transient mixed-id read (the writer Put
+        between two readers' Gets) fails here and is rejected, to be retried,
+        rather than accepted out of lockstep.
+
+        Returns True if not synchronizing, or if all ranks read the same id.
+        """
+        if not synchronize:
+            return True
+        local_val = np.array((new_id,), 'i')
+        sum_ids = np.zeros(1, 'i')
+        self.cylinder_comm.Allreduce((local_val, MPI.INT),
+                                     (sum_ids, MPI.INT),
+                                     op=MPI.SUM)
+        return new_id * self.cylinder_comm.size == sum_ids[0]
+
+    def _mark_new(self, buf: RecvArray, new_id: int) -> bool:
+        """Commit an accepted read: stamp ``new_id`` into the buffer's id slot
+        and mark it new. The data itself must already be in ``buf`` (placed by
+        the caller's single Get, or by overlap-map assembly for a multi-source
+        read). Writing the id slot before ``_pull_id`` lets the multi-source
+        path -- whose assembly fills only the data region -- reuse the same
+        commit as the single-source/equal-rank paths.
+        """
+        buf._is_new = True
+        buf._array[-1] = new_id
+        buf._pull_id()
+        return True
+
     def get_receive_buffer(self,
                            buf: RecvArray,
                            field: Field,
@@ -576,23 +612,15 @@ class SPCommunicator:
 
         new_id = int(buf.array()[-1])  # logical view
 
-        if synchronize:
-            local_val = np.array((new_id,), 'i')
-            sum_ids = np.zeros(1, 'i')
-            self.cylinder_comm.Allreduce((local_val, MPI.INT),
-                                         (sum_ids, MPI.INT),
-                                         op=MPI.SUM)
-            if new_id * self.cylinder_comm.size != sum_ids[0]:
-                buf._is_new = False
-                return False
-
-        if new_id > last_id:
-            buf._is_new = True
-            buf._pull_id()
-            return True
-        else:
+        if not self._write_ids_agree(new_id, synchronize):
             buf._is_new = False
             return False
+
+        if new_id > last_id:
+            # data is already in buf from the Get above
+            return self._mark_new(buf, new_id)
+        buf._is_new = False
+        return False
 
     # ------------------------------------------------------------------
     # Unequal-rank (Option D) helpers. These are reached only when
@@ -632,6 +660,12 @@ class SPCommunicator:
             )
         _, logical_len, _ = layout[field]
         remote_data_len = logical_len - 1  # exclude the trailing write_id slot
+        # The segment occupies the half-open remote range
+        # [remote_offset, remote_offset + count); the guard is an overrun
+        # check on its exclusive end, so '>' (ending exactly at remote_data_len
+        # is legal) and remote_offset belongs in the LHS (a segment may start
+        # mid-buffer). A reader needing only a subset of a source's scenarios
+        # legitimately ends before remote_data_len, so this is not '!='.
         if seg.remote_offset + seg.count > remote_data_len:
             raise RuntimeError(
                 f"{self.__class__.__name__}: overlap segment for {field} reads "
@@ -670,19 +704,12 @@ class SPCommunicator:
         last_id = buf.id()
         self.window.get(buf.window_array(), window_rank, field)
         new_id = int(buf.array()[-1])
-        if synchronize:
-            local_val = np.array((new_id,), 'i')
-            sum_ids = np.zeros(1, 'i')
-            self.cylinder_comm.Allreduce((local_val, MPI.INT),
-                                         (sum_ids, MPI.INT),
-                                         op=MPI.SUM)
-            if new_id * self.cylinder_comm.size != sum_ids[0]:
-                buf._is_new = False
-                return False
+        if not self._write_ids_agree(new_id, synchronize):
+            buf._is_new = False
+            return False
         if new_id > last_id:
-            buf._is_new = True
-            buf._pull_id()
-            return True
+            # data is already in buf from the Get above
+            return self._mark_new(buf, new_id)
         buf._is_new = False
         return False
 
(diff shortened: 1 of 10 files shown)


mpisppy/cylinders/spcommunicator.py
```diff
@@ -498,6 +594,163 @@ def get_receive_buffer(self,
...
+        layout = self.window.strata_buffer_layouts[remote_global_rank]
+        if field not in layout:
+            raise RuntimeError(
+                f"{self.__class__.__name__} on cylinder {self.cylinder_index} "
+                f"expected {field} on global rank {remote_global_rank} but it "
+                f"is not published there."
+            )
+        _, logical_len, _ = layout[field]
+        remote_data_len = logical_len - 1  # exclude the trailing write_id slot
+        if seg.remote_offset + seg.count > remote_data_len:
```
bknueven, line 669: Is `>` the right check? It's not `!=`?

Maybe `remote_offset` shouldn't be included in the LHS?

mpisppy/cylinders/spcommunicator.py
```diff
@@ -498,6 +594,163 @@ def get_receive_buffer(self,
...
+            self._validate_segment(field, seg.remote_rank, seg)
+        self.overlap_maps[(field, peer_cylinder)] = segments
+        self._overlap_source_ranks[(field, peer_cylinder)] = \
+            sorted({seg.remote_rank for seg in segments})
+
+    def _flex_get_single_source(self, buf, field, peer_cylinder, synchronize):
+        """Read a global/scalar field single-source from a peer cylinder's base
+        rank. Every rank of the cylinder holds the identical value, so reading
+        the base rank is sufficient and keeps the cross-cylinder write_id
+        agreement check (every local rank reads the same remote rank)."""
```
bknueven, line 700: Is this the same as the calling method?

mpisppy/cylinders/spcommunicator.py
```diff
@@ -498,6 +594,163 @@ def get_receive_buffer(self,
...
+
+        if synchronize:
+            local_val = np.array((new_id,), 'i')
+            sum_ids = np.zeros(1, 'i')
+            self.cylinder_comm.Allreduce((local_val, MPI.INT),
+                                         (sum_ids, MPI.INT),
+                                         op=MPI.SUM)
+            if new_id * self.cylinder_comm.size != sum_ids[0]:
+                buf._is_new = False
+                return False
```
bknueven: This is a pretty critical piece of logic -- let's consider factoring into a helper so if it needs to be changed it's changed everywhere.

mpisppy/cylinders/spcommunicator.py
```diff
@@ -498,6 +594,163 @@ def get_receive_buffer(self,
...
+
+        if new_id > last_id:
+            data_view = buf.value_array()
+            for seg in self.overlap_maps[(field, peer_cylinder)]:
+                snapshot = source_snapshots[seg.remote_rank]
+                data_view[seg.local_offset : seg.local_offset + seg.count] = \
+                    snapshot[seg.remote_offset : seg.remote_offset + seg.count]
+            buf._is_new = True
+            buf._id = new_id
+            buf._array[-1] = new_id
```
bknueven: No call to `_pull_id` like above?

mpisppy/cylinders/spcommunicator.py
DLWoodruff: Kept as-is — this is an overrun guard, not an equality check. A segment occupies the half-open remote range `[remote_offset, remote_offset + count)`, so the test is on its exclusive end: `>` because ending *exactly* at `remote_data_len` is legal, and `remote_offset` belongs in the LHS because a segment can start mid-buffer. It is deliberately not `!=` — a reader that needs only a subset of a source's scenarios legitimately ends before `remote_data_len`. Added a comment spelling this out in bf090982.

mpisppy/cylinders/spcommunicator.py
DLWoodruff: Yes — it was a near-copy of the equal-rank `get_receive_buffer` body. Factored the shared write_id agreement and commit into `_write_ids_agree` and `_mark_new`; both the equal-rank path and this single-source helper now call them, so they can no longer drift apart. bf090982.

mpisppy/cylinders/spcommunicator.py
DLWoodruff: Agreed — extracted into `_write_ids_agree`, now the single source of truth for the agreement check, called by `get_receive_buffer`, `_flex_get_single_source`, and `_flex_get_multi_source`. The commit half is likewise shared via `_mark_new`. bf090982.

mpisppy/cylinders/spcommunicator.py
DLWoodruff: Good catch — the asymmetry is gone. Multi-source now commits via the shared `_mark_new`, which writes the id slot (`buf._array[-1] = new_id`) then calls `_pull_id`, exactly like the single-source/equal-rank paths; the hand-set `buf._id = new_id` is removed. bf090982.

mpisppy/tests/test_spwindow_multisource.py
```diff
@@ -0,0 +1,133 @@
...
+
+        win = SPWindow(my_fields, fullcomm)
+
+        if is_writer:
+            padded = win.buffer_layout[Field.NONANTS_VALS][2]
+            buf = np.full(padded, np.nan, dtype="d")
+            for i, scen in enumerate(my_scen):
+                for j in range(k):
+                    buf[i * k + j] = _scen_value(scen, j)
+            win.put(buf, Field.NONANTS_VALS)
```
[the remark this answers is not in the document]
DLWoodruff: Done — the writer field is now sized `data_len + 1` to mirror the production trailing-id-slot layout. The overlap segments still address only the data region, so the test now additionally guards that assembly never reads into the id slot. bf090982.

run_coverage.bash
```diff
@@ -174,6 +174,14 @@ run_phase "test_dualcg_main (serial)" \
 run_phase "test_dualcg_with_cylinders (mpiexec -np 2)" \
     mpiexec -np 2 coverage run --rcfile="$PROJ_DIR/.coveragerc" -m mpi4py mpisppy/tests/test_dualcg_with_cylinders.py
 
+# Flexible (unequal) rank assignments: both need 6 ranks (4-rank + 2-rank
+# cylinders); -oversubscribe so they run on hosts with fewer than 6 cores.
+run_phase "test_spwindow_multisource (mpiexec -np 6)" \
+    mpiexec -np 6 -oversubscribe coverage run --rcfile="$PROJ_DIR/.coveragerc" -m mpi4py mpisppy/tests/test_spwindow_multisource.py
+
+run_phase "test_flexible_rank_cylinders (mpiexec -np 6)" \
+    mpiexec -np 6 -oversubscribe coverage run --rcfile="$PROJ_DIR/.coveragerc" -m mpi4py mpisppy/tests/test_flexible_rank_cylinders.py
```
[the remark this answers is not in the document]
DLWoodruff: Fixed — `-oversubscribe` is now gated behind an `mpiexec --version` probe (the `OVERSUBSCRIBE` var is set only when mpiexec reports OpenMPI/OpenRTE), so the script no longer passes an unknown flag under MPICH. bf090982.

mpisppy/tests/test_flexible_rank_ratios.py
```diff
@@ -33,36 +44,41 @@ def Get_rank(self):
...
+    def test_nonuniform_no_longer_raises_not_implemented(self):
+        # A feasible non-uniform request used to raise NotImplementedError.
+        # Now it proceeds to build communicators; on the bare stub (no Split)
+        # that surfaces as some other exception -- proving the gate is gone.
+        hub = _min_dict("hub", rank_ratio=1.0)
+        spoke = _min_dict("spoke", rank_ratio=0.5)
+        ws = WheelSpinner(hub, [spoke])
+        with self.assertRaises(Exception) as ctx:
+            ws.run(comm_world=_StubComm(size=4))
+        self.assertNotIsInstance(ctx.exception, NotImplementedError)
```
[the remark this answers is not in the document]
DLWoodruff: Narrowed to `assertRaises(AttributeError)` (the stub comm has no `Split`). Since `AttributeError` is not a `NotImplementedError`, that alone proves the gate is gone, so the extra `assertNotIsInstance` check is dropped. bf090982.

mpisppy/tests/test_flexible_rank_ratios.py
```diff
@@ -33,36 +44,41 @@ def Get_rank(self):
...
+
+    def test_uniform_ratios_use_equal_path(self):
+        # All ratios 1.0 -> equal path; reaches _make_comms and fails on the
+        # stub (no Split). Reaching that point at all confirms the unequal
+        # branch was not taken and nothing raised NotImplementedError.
         hub = _min_dict("hub", rank_ratio=1.0)
         spoke = _min_dict("spoke", rank_ratio=1.0)
         ws = WheelSpinner(hub, [spoke])
```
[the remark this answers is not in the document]
DLWoodruff: Same here — narrowed to `assertRaises(AttributeError)` and dropped the now-redundant `assertNotIsInstance(NotImplementedError)`. bf090982.

PR #735, opened 2026-05-29
stoch-admm phases C/D/E/F: naming defaults, inparser, archive

### LINKED ISSUE #734: Archive deprecated drivers in examples/stoch_distr/
With the phase-A-through-E ADMM user-API automation work landed, the canonical path for running stoch-admm is \`generic_cylinders --stoch-admm --module-name <module>\`.  Three drivers under \`examples/stoch_distr/\` predate that path and now exist mainly as historical references:

- \`stoch_distr_admm_cylinders.py\` — wires WheelSpinner directly; was the canonical entry point pre-\`generic_cylinders\`.
- \`stoch_distr_ef.py\` — depends on \`_make_admm\` from the above; intended as an EF baseline for the cylinder run.  \`generic_cylinders --EF --stoch-admm\` covers the same case.
- \`globalmodel.py\` — runs the global (non-decomposed) model; useful pedagogically but tangential to the ADMM path.

Each driver carries its own arg-registration boilerplate (e.g. \`--num-admm-subproblems\`, \`--num-stoch-scens\` post-phase-E), and changes to the user API force matching edits in all three.

Proposal: move them under \`examples/stoch_distr/archive/\` (or a similarly named subdirectory) and update \`doc/src/generic_admm.rst\` and the \`Branching factors\` worked-example \`stoch_admm_stage2ef.bash\` accordingly.  Also confirm whether anyone is using the archived drivers programmatically before moving.

Out of scope here: deleting the drivers outright.

DLWoodruff:
## Summary

- **Phase C** (`58b6623f`): Library-side naming defaults in `mpisppy.utils.stoch_admmWrapper` — `default_combining_names`, `default_split_admm_stoch_subproblem_scenario_name`, `default_admm_stoch_subproblem_scenario_names_creator` (delimiter `__ADMM__`).  `setup_stoch_admm` / `setup_stoch_admm_with_bundles` fall back to the defaults via a new `_discover_naming_helpers` (both-or-neither on the `combining_names` / `split_admm_stoch_subproblem_scenario_name` inverse pair).  `Stoch_AdmmWrapper(split=None, ...)` resolves to the default.
- **Phase D** (`bf5af9c2`): Migrate `examples/stoch_distr/stoch_distr.py` to the defaults — drops the three user-defined naming functions; `scenario_creator` decodes the composite name via the package default.
- **Phase E** (`c8596572`): Drop the redundant `num_admm_subproblems` / `num_stoch_scens` registrations from `stoch_distr.inparser_adder` (`admm_args` is the canonical source).
- **Review follow-up** (`7387f905`, addresses @emglista): the library does *not* require `--num-admm-subproblems` / `--num-stoch-scens` under `--stoch-admm`.  The wrapper derives the counts from the lengths of the name lists the model returns, so the flags are only an input to count-based name creators (like `stoch_distr`'s `range(num_...)`).  The friendly "you forgot the count" guard now lives in `stoch_distr`'s name creators, where the count-based assumption actually is; a model that builds names another way (e.g. one scenario per day across a date range) can ignore the flags entirely.
- **Phase F** (`df56bdd3`, added at the user's request — closes #734): Move the three deprecated drivers (`stoch_distr_admm_cylinders.py`, `stoch_distr_ef.py`, `globalmodel.py`) plus `go.bash` / `scale_up.bash` under `examples/stoch_distr/archive/` so new users don't mistake them for the recommended pattern.  sys.path fixups added; integration tests' EF baseline switched to `generic_cylinders.py --EF`.
- **Doc rewrite** (`fd09def5`): The "Naming the composite ADMM-stochastic scenarios" subsection in `doc/src/generic_admm.rst` now leads with *"Recommended: do not define any naming helpers on your module"* and a snippet showing the only thing users do touch (decoding the composite name in `scenario_creator`).  Customization is framed as a rare escape hatch.  Consistent rephrasing in the "Extending to Stochastic ADMM" list and the composite-name `Note`.

Refs `doc/designs/admm_user_api_automation_design.md` Phases C, D, E.

## Test plan
- [x] `ruff check .` clean.
- [x] `python -m pytest mpisppy/tests/test_stoch_admmWrapper.py mpisppy/tests/test_admm_bundler.py mpisppy/tests/test_admmWrapper.py` — 87 passed, 2 skipped (pre-existing skips).
- [x] EF integration smoke: `python ../../mpisppy/generic_cylinders.py --module-name stoch_distr --EF --EF-solver-name xpress --stoch-admm --num-admm-subproblems 3 --num-stoch-scens 2` from `examples/stoch_distr/` produces `EF objective: -86183.921722416`; same objective from `archive/stoch_distr_ef.py` after the move.
- [x] Missing-counts error: `python generic_cylinders.py --module-name stoch_distr --stoch-admm --EF --EF-solver-name xpress` raises `RuntimeError: --stoch-admm requires --num-admm-subproblems; pass it on the command line.`
- [x] `cd doc && make html` reads `generic_admm.rst` without warnings (existing unrelated build warnings remain).
- [x] CI to confirm above on the canonical solver set.


doc/src/generic_admm.rst
```diff
@@ -580,9 +609,20 @@ Argument                                    Domain       Description
 ==========================================  ===========  ============================================
 ``--admm``                                  bool         Enable deterministic ADMM decomposition
 ``--stoch-admm``                            bool         Enable stochastic ADMM decomposition
+``--num-admm-subproblems``                  int          Number of ADMM subproblems (stoch-admm only)
```
emglista, line 612: I am not using --num-admm-subproblems or --num-stoch-scenarios for my scenario creator (e.g., I give the creator a date range and split by day), and I don't think these are strictly necessary.  Might be worth removing them as required inputs since they seem specific to how the stoch_distr example is creating scenarios.

examples/stoch_distr/archive/globalmodel.py
```diff
@@ -21,14 +27,26 @@ def _parse_args():
     # create a config object and parse
     cfg = config.Config()
     stoch_distr.inparser_adder(cfg)
+    # admm subproblem / stochastic scenario counts: registered by
+    # mpisppy.generic.admm.admm_args under generic_cylinders --stoch-admm,
+    # but this non-decomposed driver runs the global model directly,
+    # so register here.
+    cfg.add_to_config("num_admm_subproblems",
```
emglista, line 34: mentioned in previous comment-- I would suggest removing these as required inputs since not every user will define scenarios/subproblems just based on a number (e.g., they may wish to pass the creator a specific set of scenarios)

emglista (commented):
Hi Dave,

I was able to get this working on my model, with the updated conventions.  I had one suggestion to make the code more generic.

Best,
Elizabeth

doc/src/generic_admm.rst
DLWoodruff: Thanks Elizabeth — you're right, and glad it's working on your model.

You've put your finger on something that was genuinely misplaced. The library never actually reads `--num-admm-subproblems` / `--num-stoch-scens`: `Stoch_AdmmWrapper` derives the counts from the *lengths* of the name lists your creators return (`len(admm_subproblem_names)`, etc.). The two flags only matter to a model whose name-creators build names from a count — which is exactly what `stoch_distr` does (`range(num_...)`). Requiring them in `_check_admm_compatibility` baked that example's assumption into the shared path and forced models like yours (names from a date range) to pass two numbers they ignore.

Fixed in `7387f905`:
- Dropped the count requirement from `_check_admm_compatibility` — `--stoch-admm` no longer needs either flag.
- Moved the friendly "you forgot the count" error into `stoch_distr`'s own name-creators, where the count-based assumption actually lives, so that example still fails clearly instead of with a `range(None)` traceback.
- Reworked the doc note to say the counts feed *your* name-creators, not the library, and that date-range-style models can skip them.

(`globalmodel.py` keeps its `required=True` — see the note on that thread.)

— Claude and Dave

examples/stoch_distr/archive/globalmodel.py
DLWoodruff: Same point as the doc thread — and fixed there for the generic library path. One exception lives here: `globalmodel.py` keeps `required=True`. It's the archived standalone driver for the count-parameterized `stoch_distr` global model (it builds the region dicts from `num_admm_subproblems` and the scenario count from `num_stoch_scens`), so it genuinely can't run without them. The change is about not imposing that requirement on every `--stoch-admm` model, not about this one example.

— Claude and Dave

PR #736, opened 2026-05-29
flex-ranks phase 3a: per-field coherence policy (DUALS strict)

DLWoodruff:
> **📚 Stacked PRs — flexible rank assignments**
> Reviewed and merged bottom-up; each PR targets the one below it.
>
> 1. **Pyomo/mpi-sppy#725 — Phase 1: infrastructure** — ✅ merged
> 2. **Pyomo/mpi-sppy#731 — Phase 2: communication layer** — ✅ merged
> 3. **Phase 3a: per-field coherence (DUALS strict)** ← _this PR_ (targets `main`)
> 4. **DLWoodruff/mpi-sppy-1#14 — Phase 3b: CLI rank-ratio flags** (targets `flex-ranks-phase3-coherence`)
> 5. **DLWoodruff/mpi-sppy-1#15 — Phase 4a: two-stage xhat fields** (targets `flex-ranks-phase3b-cli`)
>
> 6. **DLWoodruff/mpi-sppy-1#16 — Phase 4b: multistage per-node xhat sourcing** (targets `flex-ranks-phase4a-xhat-2stage`)
> 7. **DLWoodruff/mpi-sppy-1#17 — Phase 5: APH (asynchronous sender) verification** (targets `flex-ranks-phase4b-multistage`)
>
> _Later phases will be appended here as they open._

---

First slice of Phase 3 (`doc/designs/flexible_rank_assignments.md`): give the
unequal-rank multi-source reader a **per-field coherence policy**, so fields
that need a single coherent iteration (DUALS) are not stitched from mixed
iterations. Equal-rank path unchanged; this is reached only at non-default
rank ratios.

## What's here

- **`reduce_source_write_ids(source_ids, strict)`** reduces a multi-source
  read's per-source write_ids:
  - *relaxed* (default): floor over sources — the assembled value may mix
    iterations, fine for fields whose consumers re-evaluate (`NONANTS_VALS`);
  - *strict* (`_STRICT_COHERENCE_FIELDS = {DUALS}`): all sources must share one
    id, else sentinel `-1` rejects the read. `Σ pₛWₛ = 0` holds only within a
    PH iteration, so a `W` stitched across iterations is an invalid Lagrangian
    bound.
- **Collective-safe strict rejection:** the sentinel (`-1`, below any real id)
  breaks the cross-reader `Allreduce`-equality check (`_write_ids_agree`), so
  every reader rejects together — no early return before the collective, no
  deadlock.

This rebases cleanly onto the merged Phase 2: `reduce_source_write_ids` feeds
its `new_id` into the shared `_write_ids_agree` / `_mark_new` helpers that
Phase 2 introduced.

## Tests

- **`test_flex_coherence_policy.py`** (serial) pins the strict/relaxed
  reduction directly — a synchronous hub never produces mixed-id sources, so
  the MPI test alone can't distinguish the policies.
- **`test_flexible_rank_duals.py`** (np=6) — farmer PH hub + Lagrangian spoke
  at **4+2 and 2+4**; DUALS crosses cylinders (strict). Both directions agree
  and the Lagrangian bound validly lower-bounds the EF optimum.

Wired into `run_coverage.bash` + the `test-cylinders` CI job. Equal-rank
regression (`test_with_cylinders` np=2) unaffected.

## Not here (separate follow-on PRs)

- **cross_scen multi-source** — `CrossScenarioCutSpoke` has a custom
  `register_receive_fields`, an `nscen`-per-scenario `CROSS_SCENARIO_COST`
  layout, and a pre-existing `FieldLengths` inconsistency (`local²` vs
  `nscen·local`) to untangle. Too complex/risky to bundle with the coherence
  framework.
- **CLI `--<spoke>-rank-ratio` flags** end to end.

BEST_XHAT/RECENT_XHATS first-stage sourcing remains deferred to Phase 4 (next
to its FWPH consumer).

### DIFF
--- a/mpisppy/cylinders/spcommunicator.py
+++ b/mpisppy/cylinders/spcommunicator.py
@@ -52,6 +52,41 @@ _GLOBAL_OR_SCALAR_FIELDS = frozenset((
     Field.EXPECTED_REDUCED_COST,
 ))
 
+# Per-scenario fields that require *strict* coherence when assembled across
+# ranks: every contributing source must be at the same write_id, or the read
+# is rejected and retried. DUALS carries the PH multipliers W_s, whose
+# normalization sum_s p_s W_s = 0 holds only within a single iteration --
+# stitching W from mixed iterations yields an invalid Lagrangian bound. Every
+# other per-scenario field uses *relaxed* coherence (the assembled value may
+# mix iterations; its consumers re-evaluate, so it is always honest).
+_STRICT_COHERENCE_FIELDS = frozenset((
+    Field.DUALS,
+))
+
+
+def reduce_source_write_ids(source_ids, strict: bool) -> int:
+    """Reduce the per-source write_ids of a multi-source read to the single id
+    this rank will compare against its last-accepted id.
+
+    Args:
+        source_ids: write_ids read from each contributing remote rank.
+        strict: if True, the sources must agree on one id (strict coherence);
+            a disagreement returns the sentinel ``-1`` so the read is rejected
+            (it is < any real id, and it breaks the cross-reader agreement
+            check, so every reader rejects together). If False (relaxed), the
+            floor over sources is taken -- a mixed-iteration assembly is
+            accepted, which is fine for fields whose consumers re-evaluate.
+
+    Returns:
+        int: the accepted write_id, or ``-1`` to reject (strict mismatch),
+        or ``0`` when there are no sources.
+    """
+    if not source_ids:
+        return 0
+    if strict:
+        return source_ids[0] if len(set(source_ids)) == 1 else -1
+    return min(source_ids)
+
 def communicator_array(data_length: int):
     """
     Allocate an MPI memory region with a padded length (multiple of 8 doubles = 64B),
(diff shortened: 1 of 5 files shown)


mpisppy/cylinders/spcommunicator.py
```diff
@@ -52,6 +52,41 @@
...
+        source_ids: write_ids read from each contributing remote rank.
+        strict: if True, the sources must agree on one id (strict coherence);
+            a disagreement returns the sentinel ``-1`` so the read is rejected
+            (it is < any real id, and it breaks the cross-reader agreement
+            check, so every reader rejects together). If False (relaxed), the
+            floor over sources is taken -- a mixed-iteration assembly is
+            accepted, which is fine for fields whose consumers re-evaluate.
+
+    Returns:
+        int: the accepted write_id, or ``-1`` to reject (strict mismatch),
```
bknueven, line 81: This is the "time to die" sentinel as well, but I guess re-use here doesn't hurt anything.

bknueven (approved):

mpisppy/cylinders/spcommunicator.py
DLWoodruff: I agree.

Issue #730, opened 2026-05-24
Flexible ranks: add CG-hub XFEAS integration test + --ph-xfeas-spoke-rank-ratio (blocked on #729)


DLWoodruff:
Follow-up to the flexible (unequal) rank assignments work. Phase 4a added
multi-source assembly for the two-stage xhat fields (`BEST_XHAT`,
`RECENT_XHATS`, `XFEAS`) and verified it end-to-end with an FWPH hub plus a
serial exact-value assembly test. Two `XFEAS`-specific items were intentionally
deferred and should be picked up once #729 is fixed.

### Blocked on #729

`XFEAS`'s only consumer is the CG hub, and a CG hub with more than one rank
deadlocks at convergence (#729). So the end-to-end `XFEAS` path at unequal ranks
cannot be exercised until #729 lands.

### To do once #729 is fixed

1. **CG-hub XFEAS integration test.** Add an np=6 test (mirroring
   `test_flexible_rank_xhat.py`): a `CGHub` + `ph_xfeas` spoke at unequal ranks
   (4+2 and 2+4), asserting both directions complete with finite, validly
   bracketing bounds near the EF optimum. Note CG needs a **non-persistent**
   master solver (`get_solver(persistent_OK=False)`) — the persistent interface
   raises "Please use set_instance ..." on the CG master.

2. **Expose `--ph-xfeas-spoke-rank-ratio`.** The CLI flag was implemented in
   phase 4a but reverted before merge, because exposing a rank ratio whose only
   consumer (the CG hub) deadlocks would be a footgun. Re-add it: a float
   defaulting to 1.0 declared in `config.py::ph_xfeas_spoke_args`, injected onto
   the spoke dict's `rank_ratio` in `generic/spokes.py::build_spoke_list`, and
   covered in `test_flexible_rank_cli.py` (matching the phase-3b flags).

The `XFEAS` multi-source assembly itself already works and is unit-tested
(`test_flex_xhat_assembly.py`); this is only the end-to-end CG coverage and the
CLI exposure that ride on the #729 fix.

DLWoodruff:
Heads up while picking this up: as written, this is blocked on more than #729.

The #729 deadlock fix is in #737 (broadcast `best_solution_obj_val` across the hub `cylinder_comm`; covered by a new `np=4` equal-rank multi-rank-CG-hub test). That clears the *deadlock*, but the two items here additionally depend on flex-ranks infrastructure that is **not yet on `main`**:

[...]

PR #737, opened 2026-05-29
fix(cg): broadcast best_solution_obj_val so a multi-rank CG hub doesn't deadlock (#729)

### LINKED ISSUE #729: CG hub deadlocks at convergence when the hub has more than one rank (best_solution_obj_val not broadcast)
### Summary

A `CGHub` run deadlocks during the convergence-path shutdown whenever the **CG hub cylinder has more than one MPI rank**. It hangs forever after printing `Terminating based on inter-cylinder relative gap ...`.

This is latent today because CG is only exercised with one rank per cylinder (the `test_cg_with_cylinders` / `test_dualcg_with_cylinders` suites run at `np=2`, i.e. a 1-rank hub + 1-rank spoke). It surfaced while bringing up flexible (unequal) rank assignments, which is the first configuration to give the CG hub more than one rank.

### Root cause

In `mpisppy/cgbase.py::iterk_loop`, the per-iteration results that feed the convergence test are computed only on `cylinder_rank == 0`:

- `best_bound_obj_val` is computed on rank 0 **and broadcast** (`cgbase.py:687`).
- `best_solution_obj_val` is computed on rank 0 but is **not** broadcast.

`CGHub.is_converged()` (`hub.py`) then does:

```python
if self.opt.best_solution_obj_val is not None:
    self.BestInnerBound = self.InnerBoundUpdate(self.opt.best_solution_obj_val)
...
terminate = self.determine_termination(...)   # uses compute_gaps() -> BestInnerBound
```

On rank 0 `best_solution_obj_val` is set, so `BestInnerBound` (hence the gap) updates; on the other hub ranks it stays `None`, so the gap differs. `determine_termination` therefore returns **different verdicts on different hub ranks**: one rank hits `break` (cgbase.py:692) and leaves `main`, while another proceeds into the next iteration and blocks at the master-problem broadcast (`cgbase.py:638`).

### Evidence

`faulthandler` stack dump of a hung 2-rank-CG-hub run: one hub rank parked at `cgbase.py:638` (`self.mpicomm.bcast(self.pi_values, root=0)`) inside `iterk_loop`, every other rank already waiting post-`main` at `spin_the_wheel.py:187`.

### Reproduce

Any CG hub with >1 rank. Easiest via flexible ranks (once that lands), e.g. a CG hub + ph_xfeas spoke at unequal ranks; it also reproduces at equal ranks with a CG hub of >1 rank (e.g. `np=4`, 2-rank hub + 2-rank spoke) on an MPI build that allows windows on a split sub-communicator.

### Likely fix

Broadcast `best_solution_obj_val` across the hub's `cylinder_comm` alongside `best_bound_obj_val` and `conv` (cgbase.py:687-688), so `is_converged()` is coherent across all hub ranks. Verify `best_solution_obj_val` is not assumed rank-0-only anywhere else, and add a multi-rank-CG-hub test (np>=4, or a flexible-rank CG configuration) so the path is covered.

DLWoodruff:
Fixes #729.

### Problem

A `CGHub` run deadlocks during convergence-path shutdown whenever the CG hub cylinder has more than one MPI rank. It hangs after printing the inter-cylinder-gap termination message.

In `CGBase.iterk_loop`, the per-iteration `best_bound_obj_val` and `conv` are broadcast across the hub's `cylinder_comm`, but `best_solution_obj_val` (set on rank 0 in the LP branch, `cgbase.py:665`) is **not**. `CGHub.is_converged()` derives `BestInnerBound` — and thus the gap that `determine_termination` checks — from `best_solution_obj_val`. So with >1 hub rank, rank 0 saw the updated gap and `break`-ed out of the iteration loop while the other ranks kept `None`, did not terminate, and blocked forever at the next master-problem broadcast (`cgbase.py:638`).

This was latent because CG is otherwise only exercised at one rank per cylinder (the `np=2` suites). It surfaces as soon as the CG hub spans more than one rank.

### Fix

Broadcast `best_solution_obj_val` across the hub's `cylinder_comm` alongside `best_bound_obj_val` and `conv`, so `is_converged()` is coherent across all hub ranks. The same defensive broadcast is added to `DCG.iterk_loop` to keep the two loops parallel (DCG does not set `best_solution_obj_val` today, so it is coherently `None` there, but this guards against a future LP branch reintroducing the divergence).

### Test

Adds `mpisppy/tests/test_cg_multirank_hub.py`: an `np=4` run that gives the CG hub two ranks (equal-rank 2+2: CG hub + xhatshuffle spoke). It converges to the EF optimum with the fix and **deadlocks without it** (verified by reverting the broadcast — the run times out). This needs no flexible-rank infrastructure; it reproduces the bug on the plain equal-rank path.

Wired into `run_coverage.bash` and `.github/workflows/test_pr_and_main.yml` (`mpiexec -np 4`).

### Note on #730

#730 (the unequal-rank XFEAS end-to-end test + `--ph-xfeas-spoke-rank-ratio`) is a separate follow-up. It is gated not only on this fix but also on the XFEAS multi-source assembly (flex-ranks phase 4a) and the CLI rank-ratio flags (phase 3b), which are not yet on `main` — on `main`, `_items_per_scen_for_field` raises `NotImplementedError` for `XFEAS`. So #730 stays open pending those phases.

### DIFF
--- a/.github/workflows/test_pr_and_main.yml
+++ b/.github/workflows/test_pr_and_main.yml
@@ -1018,6 +1018,12 @@ jobs:
           cd mpisppy/tests
           mpiexec -np 2 coverage run $COV_ARGS -m mpi4py test_cg_with_cylinders.py
 
+      - name: run multi-rank CG hub test
+        timeout-minutes: 10
+        run: |
+          cd mpisppy/tests
+          mpiexec -np 4 coverage run $COV_ARGS -m mpi4py test_cg_multirank_hub.py
+
       - name: run DCG serial tests
         timeout-minutes: 10
         run: |
--- a/mpisppy/cgbase.py
+++ b/mpisppy/cgbase.py
@@ -683,10 +683,15 @@ class CGBase(mpisppy.spopt.SPOpt):
                     print("Iteration time: %6.2f" % (time.time() - iteration_start_time))
                     print("Elapsed time:   %6.2f" % (time.perf_counter() - self.start_time))
 
-            # Broadcast updated bound and convergence metric
+            # Broadcast updated bounds and convergence metric so that
+            # is_converged() (which feeds determine_termination via the inner
+            # bound) reaches the same verdict on every hub rank. Without
+            # broadcasting best_solution_obj_val, only rank 0's gap updates and
+            # ranks can disagree on termination, deadlocking a multi-rank hub.
             self.best_bound_obj_val = self.mpicomm.bcast(self.best_bound_obj_val, root=0)
+            self.best_solution_obj_val = self.mpicomm.bcast(self.best_solution_obj_val, root=0)
             self.conv = self.mpicomm.bcast(self.conv, root=0)
-            
+
             if self.spcomm and self.spcomm.is_converged():
                 global_toc("Cylinder convergence", self.cylinder_rank == 0)
                 break
--- a/mpisppy/opt/dualcg.py
+++ b/mpisppy/opt/dualcg.py
@@ -267,7 +267,10 @@ class DCG(mpisppy.cgbase.CGBase):
                     print("Iteration time: %6.2f" % (time.time() - iteration_start_time))
                     print("Elapsed time:   %6.2f" % (time.perf_counter() - self.start_time))
                 
+            # Broadcast both bounds and the convergence metric so is_converged()
+            # reaches the same verdict on every hub rank (see CGBase.iterk_loop).
             self.best_bound_obj_val = self.mpicomm.bcast(self.best_bound_obj_val, root=0)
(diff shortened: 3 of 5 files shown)


bknueven (approved):
I assume rank 0 "owns" the objective value?

DLWoodruff:
Yes — rank 0 owns it. The restricted master is solved only under `if self.cylinder_rank == 0:` in `iterk_loop`, and `best_solution_obj_val = self.rmp_obj_val` is set inside that block, so rank 0 is the sole writer. That's exactly the bug: the value was computed on rank 0 but never shared, so the other hub ranks kept a stale gap and disagreed on `is_converged()`. The fix just adds it to the `root=0` broadcast next to the `best_bound_obj_val` and `conv` broadcasts that were already there (and mirrors the same one line in `dualcg.py`).

PR #739, opened 2026-05-30
Cap solver threads in the test suite (#718)

### LINKED ISSUE #718: Tests don't constrain solver thread count
Most pytest test cases instantiate a Pyomo solver (cplex / gurobi / xpress) and leave the thread count at the solver's default — which on modern CPUs means each solve fans out to all physical cores. With many tests running in sequence (and several pytest processes in CI), this saturates the host, slows the suite, and makes test timings noisy on shared runners.

Suggested fix: set `Threads=1` (or the per-solver equivalent) wherever tests construct a solver — most centrally in `mpisppy/tests/utils.py::get_solver`.

Out of scope: production code paths (those should keep solver-thread tuning configurable; this is only about the test environment).

DLWoodruff:
Fixes #718.

## Problem
Pytest test cases instantiated a Pyomo solver (cplex/gurobi/xpress) and left the thread count at the solver's default, so on modern CPUs each solve fanned out to all physical cores. With many tests running in sequence (and several pytest processes in CI), this saturated shared hosts, slowed the suite, and made timings noisy.

## Fix (test-only; production code paths untouched)
Cap every solver built by the **test suite** at one thread, reusing existing machinery:

- **Direct `SolverFactory().solve()` sites** — new `tests/utils.py::limit_solver_threads(solver, solver_name, threads=1)` helper that sets the solver-native thread option via the existing `sputils.translate_solver_options` translator (no hardcoded per-solver keys). Applied in `test_ef_ph`, `test_cg_main`, `test_dualcg_main`, `test_feasible_xhat`, `test_mps`, `test_proper_bundler`, and the `ExtensiveForm` path in `test_pysp_model`.
- **Subprocess CLI runs** — added the existing `--max-solver-threads=1` flag to `generic_cylinders` / admm-cylinders commands that lacked it (`straight_tests`, `test_admmWrapper`, `test_stoch_admmWrapper`, `test_pre_pickle_pipeline`, `test_generic_cylinders`).
- **In-process framework solves** — injected canonical `"threads": 1` into iter0/iterk solver-option dicts, or set `cfg.max_solver_threads = 1` (`test_ef_ph`, `test_aph`, `test_ph_extensions`, `test_ph_main`, `test_agnostic`, `test_gradient_rho`, `test_xbar_w_reader_writer`).

Existing `--max-solver-threads=2` commands are left unchanged.

## Known limitation
The `distr_ef.py` / `stoch_distr_ef.py` EF-comparison solves invoked by the ADMM tests stay uncapped — they are production example scripts without `popular_args()`, so they don't accept `--max-solver-threads` (out of scope for a test-only change).

## Verification
`ruff check` clean; `git diff` touches only `mpisppy/tests/`. Ran the affected serial suites locally (xpress): direct-solve group (46 passed), PH/APH group (21 passed), gradient/pysp group passed, and the fast `test_ef_ph` cases including the edited direct-solve tests. Two unrelated, pre-existing failures on the local xpress setup were confirmed to reproduce **without** this change and are not introduced here (`test_xbar_w_reader_writer::test_xbarwriter` hard-coded-value mismatch; `test_ef_ph::Test_sizes::test_persistent_basic` slowness).


mpisppy/tests/test_pre_pickle_pipeline.py
```diff
@@ -331,6 +331,7 @@ def test_pickle_scenarios_then_solve_with_all_stages(self):
                 os.chdir(self._farmer_dir())
                 cmd_pickle = (
                     f"{python} {python_args} -m mpisppy.generic_cylinders "
+                    f"--max-solver-threads 1 "
```
bknueven, line 335: Based on the test failures, it seems like `--module-name` needs to be the first argument? But why would that be since it's named?

DLWoodruff: You raise a good point.  This is a hold-over from before it was named. Now that it is named, "we" can just scan the input buffer for --module-name "by hand" and then pull out the name. I will ask Claude to do that. That will be another PR, but a very small one.

PR #746, opened 2026-06-11
CVaR Phase 1: core risk-management transform + EF tests

DLWoodruff:
## What

Phase 1 of CVaR (Conditional Value-at-Risk) risk management, per the design doc included here (`doc/designs/cvar_design.md`). Supports minimizing a risk-averse objective `λ·E[Cost] + β·CVaR_α(Cost)` (PySP-style weighted CVaR).

This phase is **programmatic-API only** — config/CLI/`generic_cylinders` wiring and a worked example come in Phase 2; maximize support and the user docs come in Phase 3.

## Approach

CVaR is implemented as a single **per-scenario model transform** (Rockafellar–Uryasev linearization), not as new hub/spoke logic. New `mpisppy/utils/cvar.py`:

- `add_cvar(scenario, *, cvar_weight, cvar_alpha, cvar_mean_weight=1.0)` adds
  - a VaR variable `η` as a **root-node nonant** (appended to both the root node's `nonant_list` and its pre-built `nonant_vardata_list`),
  - a scenario-local excess `δ_s ≥ Cost_s − η` (`δ_s ≥ 0`),
  - deactivates the original risk-neutral objective (left on the model for E[Cost] reporting) and adds a new active objective `WITH_CVAR = λ·Cost + β·η + β/(1−α)·δ_s`.
- `cvar_scenario_creator(scenario_creator, ...)` wraps any creator to apply the transform.

Because `Σ_s p_s = 1` and `η` is one shared first-stage variable, the risk measure distributes over scenarios and `η` becomes *"just another first-stage variable."* Every cylinder then inherits risk aversion with **no algorithm changes**, and the EF reproduces PySP's EF-CVaR exactly. This is robust because all mpi-sppy objective lookups filter on `active=True`, so the deactivated original is invisible and only `WITH_CVAR` is used.

Two deliberate PySP deviations (see design §6): no `weight==0 ⇒ pure CVaR` overload (pure CVaR is `--cvar-mean-weight 0`), and we drop PySP's stray constant initializer. **Scope note:** a single root-stage `η` over total cost — the same formulation as PySP — which is *not* a nested/time-consistent multistage measure (documented in design §6.5; ships verbatim in the Phase 3 docs).

## Tests (`mpisppy/tests/test_cvar.py`, 10 tests)

- **Structural** (no solver): `WITH_CVAR` is the sole active objective, original deactivated-but-present, `η` appended to root nonants, excess var/constraint present, validation guards raise.
- **EF-CVaR vs closed form** on a tiny deterministic instance (costs {10,20,30,40} uniform, α=0.6 ⇒ E[Cost]=25, VaR η\*=30, CVaR=36.25): checks the objective and η.
- **Pure CVaR** (λ=0) and the **β=0 risk-neutral regression** reproducing the plain farmer EF objective.

`test_cvar` is wired into `run_coverage.bash` and `.github/workflows/test_pr_and_main.yml` in the same commit so codecov/patch covers `cvar.py` on this PR.


mpisppy/utils/cvar.py
```diff
@@ -0,0 +1,116 @@
...
+        raise ValueError(f"cvar_alpha must satisfy 0 < alpha < 1 (got {cvar_alpha})")
+    if cvar_weight < 0.0:
+        raise ValueError(f"cvar_weight (beta) must be >= 0 (got {cvar_weight})")
+    if cvar_mean_weight < 0.0:
+        raise ValueError(
+            f"cvar_mean_weight (lambda) must be >= 0 (got {cvar_mean_weight})")
+
+    obj = sputils.find_active_objective(scenario)   # pristine user cost objective
+    cost = obj.expr
+    sense = obj.sense                               # minimize; maximize is TBD
```
bknueven: We should either handle maximize explicitly (preferred) or throw an error for a maximization problem.

bknueven (approved):
Approved pending fix for maximization

mpisppy/utils/cvar.py
DLWoodruff: Done in 99a2b480: `add_cvar` now raises `NotImplementedError` for a maximization objective rather than silently building the upper-tail formulation, since this transform is only correct for minimization. Added a structural test (`test_maximize_objective_raises`) and synced the design doc. Full maximize support (the mirrored lower-tail formulation) remains scheduled for Phase 3.

PR #755, opened 2026-06-14
Add some tests for sep_rho, prox_approx, ciutils, reduced_costs_fixer

DLWoodruff:
## What

Adds tests for deterministic logic that was under-covered. Two parts:

1. **Solver-free, MPI-free unit tests** for four modules, wired into the CI no-solver unit-test job and `run_coverage.bash`.
2. **A solver-gated end-to-end test** that PH with the linearized proximal term agrees with PH using the exact quadratic prox, wired into the cplex/xpress regression job and `run_coverage.bash`.

## Unit tests added

| Module | Coverage | What the tests verify |
|---|---|---|
| `extensions/sep_rho.py` | 21% → 92% | the SEP rho formula — integer `\|c\|/(xmax−xmin+1)` vs continuous `\|c\|/max(1,E\|x−xbar\|)`, abs of a negative cost, the `max(1,·)` clamp, multiplier scaling, zero-cost skip — and the cross-scenario xmax/xmin/residual all-reduces |
| `extensions/reduced_costs_fixer.py` | 42% → 67% | the fix/unfix heuristic — fix-to-bound by reduced-cost sign and bound proximity (both senses), the quantile target cutoff, unfix on rc decay / xbar drift / nan, and the modeler-fixed / surrogate / all-nan guards |
| `utils/prox_approx.py` | no direct test before (≈50% indirect, continuous path only, via `test_proper_bundler`) | tolerance-aware cut insertion at every bisect branch, tangent/secant cut geometry (tight at the lattice points, lower-bounding off them), bound clamping, dedup — and, importantly, the **discrete/integer** manager and `_compute_mb`, which farmer (continuous) never exercised |
| `confidence_intervals/ciutils.py` | partial before | prime factorization, branching-factor construction/scaling, node-list ordering validation, sign-aware numeric gap correction |

All four run as fast serial unit tests (no MPI, no solver). They check actual computed values against the formulas, not just that the code runs.

## End-to-end test added

`tests/test_prox_approx_e2e.py` drives the real PH engine on farmer-3. The unit tests above call `prox_approx` with `persistent_solver=None` and never solve a subproblem, so they verify cut *geometry* but not the *consequence*: that linearizing the prox term steers PH to the right answer. This test runs PH twice — `linearize_proximal_terms=True` vs the exact quadratic prox — and asserts:

- the converged first-stage solution matches between the two runs (within 1.0 acre; observed ≈0.01),
- both runs match the extensive-form optimum (so a shared-but-wrong fixed point can't pass),
- the reported objectives agree and recover the EF optimum.

It is solver-gated (`@skipUnless`); the exact-prox run needs a QP-capable solver, and `get_solver` only returns cplex/gurobi/xpress, all of which qualify.

## Harness

- The four unit-test files are added to the CI `run unit tests` (no-solver) job and to `run_coverage.bash`.
- `test_prox_approx_e2e.py` is added to the **regression job** (which installs cplex + xpress) and to `run_coverage.bash` — not the no-solver job, where the QP run would be skipped and contribute nothing.
- `run_coverage.bash` also gains the serial unit-test block (`test_sputils`, `test_config`, `test_log`, `test_scenario_tree`, `test_nice_join`, `test_solver_spec`, `test_extensions`, `test_buffer_inspect`, `test_comm_lor_check`) that CI already runs under coverage but the local harness was missing — so local coverage numbers now match CI.

### DIFF
--- a/.github/workflows/test_pr_and_main.yml
+++ b/.github/workflows/test_pr_and_main.yml
@@ -1155,7 +1155,11 @@ jobs:
             mpisppy/tests/test_flexible_rank_ratios.py \
             mpisppy/tests/test_flex_coherence_policy.py \
             mpisppy/tests/test_flexible_rank_cli.py \
-            mpisppy/tests/test_flex_xhat_assembly.py
+            mpisppy/tests/test_flex_xhat_assembly.py \
+            mpisppy/tests/test_ciutils.py \
+            mpisppy/tests/test_prox_approx.py \
+            mpisppy/tests/test_sep_rho.py \
+            mpisppy/tests/test_reduced_costs_fixer.py
 
       - name: Upload coverage data
         if: always()
--- a/.github/workflows/test_pr_and_main.yml
+++ b/.github/workflows/test_pr_and_main.yml
@@ -124,6 +124,10 @@ jobs:
         run: |
           coverage run $COV_ARGS -m pytest mpisppy/tests/test_feasible_xhat.py -v
 
+      - name: Test prox_approx end-to-end
+        run: |
+          coverage run $COV_ARGS -m pytest mpisppy/tests/test_prox_approx_e2e.py -v
+
       - name: Test docs
         run: |
           cd ./doc/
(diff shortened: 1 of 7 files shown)


bknueven (approved):
I only read the prox approx test thoroughly, and that one is pretty good!

PR #757, opened 2026-06-14
Flex ranks phase 7: reduced_costs rank-ratio flag (equal-rank only for the deprecated reduced-cost rho)

DLWoodruff:
> **Related PRs — flexible-rank reduced-costs thread.** ✅ #758 (extension-field multi-source fix) is **MERGED**. Two PRs remain off `main`; recommended merge order:
> 1. **#759** — deprecate & remove `reduced_costs_rho`
> 2. **#757** — flex phase 7: `--reduced-costs-rank-ratio` flag
>
> After #759 lands, `reduced_costs_rho` always errors, so #757's `checker()` guard for it becomes dead code — when rebasing #757 last, drop that guard (its `--reduced-costs-rank-ratio` flag stays useful). #757 and #759 both edit `Config.checker()`, so expect a conflict there.

---

Completes the per-spoke rank-ratio flag coverage for flexible (unequal) rank
assignments by exposing `--reduced-costs-rank-ratio`.

### What works at unequal ranks

The reduced-costs spoke's flex-relevant fields already have what they need:

- the Lagrangian outer bound (`DUALS`) and the nonants it receives from the hub
  (`NONANTS_VALS`) have multi-source assemblers from earlier phases;
- the reduced-cost **fixer** reads `EXPECTED_REDUCED_COST`, a global field read
  single-source.

So the spoke runs correctly at unequal ranks for those uses, and this PR exposes
the flag for them — one `add_to_config` in `reduced_costs_args` plus one
`build_spoke_list` line, mirroring the phase-6 flags. The equal-rank path is
untouched.

### The one unsupported case: reduced-cost rho

`SCENARIO_REDUCED_COST` is the only per-scenario field whose **sole** consumer is
`ReducedCostsRho` — which is itself scheduled for deprecation (reduced-cost rho
has not proven effective in practice, #673). Rather than build out a multi-source
assembler for a deprecated-only field, it is deliberately left without one.

To keep the new flag from *silently* mis-assembling that field, `Config.checker()`
rejects the single unsupported combination — `--reduced-costs-rho` together with a
non-1.0 `--reduced-costs-rank-ratio` — with an actionable message (run reduced-cost
rho at equal ranks). If the deprecation is ever reversed, the remaining work is the
small `XFEAS`→CG-hub-style addition noted in the design doc.

### Tests

`test_flexible_rank_cli.py` covers the `reduced_costs_rank_ratio` flag (default and
`build_spoke_list` injection) and the `checker()` guard (rejected at unequal ranks,
allowed at equal ranks).

`ruff` clean; serial flex tests green; the existing `np=6` flex MPI tests and the
`np=2` equal-rank regression are unaffected (no communicator-layer changes in this
PR).

The per-spoke flag coverage table in `doc/designs/flexible_rank_assignments.md` is
now complete; `SCENARIO_REDUCED_COST` is the only per-scenario field still without a
multi-source assembler, by deliberate choice (deprecated consumer).

### DIFF
--- a/doc/designs/flexible_rank_assignments.md
+++ b/doc/designs/flexible_rank_assignments.md
@@ -779,7 +779,12 @@ Coverage therefore tracks field support, not spoke convenience:
 | `relaxed_ph` | `relaxed_ph_rank_ratio` | `DUALS`, `RELAXED_NONANTS_VALS` | landed |
 | `subgradient` | `subgradient_rank_ratio` | `XFEAS`; `NONANTS_VALS` (recv) | landed |
 | `fwph` | `fwph_rank_ratio` | `BEST_XHAT` / `RECENT_XHATS` (recv); `NONANTS_VALS` (recv) | landed |
-| `reduced_costs` | — | `SCENARIO_REDUCED_COST` | needs assembler + consumer rework — see phase 7 |
+| `reduced_costs` | `reduced_costs_rank_ratio` | `DUALS`, `NONANTS_VALS` (recv); `SCENARIO_REDUCED_COST`† | flag landed (phase 7) |
+
+† `SCENARIO_REDUCED_COST`'s only consumer is the deprecated `ReducedCostsRho`,
+so it was left without a multi-source assembler (see Phase 7). The spoke's
+other uses work at unequal ranks; `checker()` rejects the one unsupported
+combination (`--reduced-costs-rho` with a non-1.0 `reduced_costs_rank_ratio`).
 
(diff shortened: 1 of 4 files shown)


DLWoodruff:
Closing as moot. #759 (merged as ca1984ad) deprecates and removes `reduced_costs_rho` entirely. This PR only added a flexible-rank rank-ratio flag for `reduced_costs_rho`, so it now configures a rho setter that no longer exists. Branch deleted.

PR #761, opened 2026-06-15
Adding (basic) LShaped support to generic cylinders

tvalenciaz:
Adding an option to run an `--lshaped-hub` and `--xhatlshaped` xhatter to generic cylinders.
Also adding an example to the `examples` based on the tutorial under development.
A test was added to the `generic_tester.py` on farmer and the new "relief" example.


doc/src/generic_cylinders.rst
```diff
@@ -124,9 +125,33 @@ Spokes provide bounds and heuristic solutions. Enable them with flags:
...
 See :ref:`Spokes` for details on each spoke type.
 
+Running L-Shaped with Spokes
+-----------------------------
+
+To run L-shaped (Benders decomposition) with bound-computing spokes for
+two-stage problems:
+
+.. code-block:: bash
+
```
DLWoodruff, line 139: I don't understand why np 3 works here. It looks to me like there are two cylinders and you have not requested unbalanced ranks. But I guess it does work. In any event, there should be a test for this particular command, which could be in the run_all part 1 script or somewhere else.

DLWoodruff (approved):

doc/src/generic_cylinders.rst
tvalenciaz: You are right, that was not on purpose; it should have been `-np 2`. I'll correct this now before merging.

PR #765, opened 2026-06-17
Slamming: preference-driven in-hub variable fixing

DLWoodruff:
## What

Adds **slamming** to mpi-sppy: forcing (fixing) selected nonanticipative
variables according to pre-specified user preferences while a decomposition hub
is running, to drive toward a feasible/converged incumbent when ordinary
convergence is slow. It is an `Extension` (in the in-hub-fixing family with
`fixer.py`, `reduced_costs_fixer.py`, `relaxed_ph_fixer.py`) and runs under any
`PHBase`-derived hub (PH, APH, Subgradient, FWPH).

This is distinct from the existing `SlamMin`/`SlamMax` *spokes*, which are
non-destructive incumbent finders; the slammer alters the hub search per user
preference and is multistage-capable (the spokes are two-stage only).

Design: `doc/designs/slamming_design.md`.

## How it's used

Activated only when a directives file is supplied, so a run with no slamming
options is byte-for-byte unchanged:

- `--slamming-directives-file <file>` — activates the extension
- `--slam-start-iter K` — first iteration at which slamming may occur (default 1)
- `--iters-between-slams M` — cadence once started (default 1)

Supplying the other options without the file is a hard error.

The directives file is a CSV keyed by nonant name with shell-style wildcards
(`*`/`?`; index brackets `[ ]` are matched literally — deliberately not
`fnmatch`, whose bracket semantics would make `DoBuild[*]` fail to match
`DoBuild[Seattle]`). Each row gives a name pattern, an optional `can_slam`
(`1`/`0`), an ordered `|`-separated list of `directions` from
`{lb, ub, nearest, anywhere, min, max}` (first applicable is used), and a
`priority` (largest is slammed first; ties by name). Matching is
last-match-wins; only variables matched by a `can_slam` rule are ever slammed;
a pattern that matches nothing is an error that names the file.

A worked example translated from PySP's `wwph.suffixes` ships at
`examples/sizes/config/slamming_directives.csv`.

## Behavior

- One variable is slammed per event, by largest priority (name-tiebroken so the
  choice is identical on every rank). Eligibility: matched + `can_slam`, not
  already fixed, not a surrogate, has an applicable direction. There is no
  convergence gate (slamming an already-settled variable just pins it).
- `lb`/`ub`/`nearest`/`anywhere` use only rank-identical data (no
  communication); `min`/`max` do one small reduction over the per-node
  communicator, and only when such a slam actually fires.
- Slams are sticky.

## Scope

This is the slamming *mechanism*. Deciding *when* to slam beyond the built-in
iteration-count trigger (genuine stall/cycle detection) is a separate future
project that would invoke this mechanism as a consumer; it is intentionally not
part of this PR.

## Tests / docs

- `mpisppy/tests/test_slammer.py` — 43 solver-free unit tests (parser, matcher,
  every direction, selection / eligibility / stickiness, trigger,
  backward-compatibility, zero-match error). Wired into `run_coverage.bash`
  **and** `test_pr_and_main.yml`.
- Docs: `doc/src/extensions.rst` (slammer section + flag) and the design doc.
- Verified live on the sizes MIP (serial and `mpiexec -np 2`): slams fire on
  cadence, the cross-rank `max` reduction stays coherent, persistent-solver
  updates work, and a bad pattern errors naming the file.

### DIFF
--- a/.github/workflows/test_pr_and_main.yml
+++ b/.github/workflows/test_pr_and_main.yml
@@ -1174,7 +1174,8 @@ jobs:
             mpisppy/tests/test_ciutils.py \
             mpisppy/tests/test_prox_approx.py \
             mpisppy/tests/test_sep_rho.py \
-            mpisppy/tests/test_reduced_costs_fixer.py
+            mpisppy/tests/test_reduced_costs_fixer.py \
+            mpisppy/tests/test_slammer.py
 
       - name: Upload coverage data
         if: always()
--- a/doc/designs/slamming_design.md
+++ b/doc/designs/slamming_design.md
@@ -1,7 +1,9 @@
 # Slamming — design
 
-**Status:** Draft — up for discussion. Nothing is implemented yet.
-Intended branch: `slamming` (off Pyomo/mpi-sppy `main`).
+**Status:** The mechanism described here is implemented on branch `slamming`
+(off Pyomo/mpi-sppy `main`). This document describes the slamming *mechanism*;
+the work that decides *when* to invoke it (stall/cycle detection) is a separate
+future project that uses slamming as a consumer — not a later phase of this one.
 **Author:** dlw (captured with Claude Code assistance)
 **Last updated:** 2026-06-17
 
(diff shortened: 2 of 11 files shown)


bknueven (approved):
I didn't find any issues in the core slamming logic. I didn't read the parser or the tests closely.

PR #767, opened 2026-06-18
Consistent rho > 0 enforcement; report (not silently inflate) zero rhos (#560)

### LINKED ISSUE #560: Consistent/rigorous enforcement of rho > 0 checks
We also should not silently inflate rhos to some small value (e.g., 0.1) if 0 values are encountered.

DLWoodruff:
## Issue

Closes #560 — "Consistent/rigorous enforcement of rho > 0 checks" and "should not silently inflate rhos to some small value (e.g., 0.1) if 0 values are encountered."

## Background

Progressive Hedging requires a strictly positive rho for **every** non-anticipative variable: rho multiplies the proximal/consensus term `(rho/2)*(x - xbar)**2`, so a non-positive value silently disables (or, if negative, inverts) enforcement of non-anticipativity. There was no central check, and several rho-setting heuristics silently substituted a value when they computed (near-)zero.

Note that a **zero objective coefficient** does **not** justify a zero rho: rho is on the consensus term, not the objective, so such a variable still needs `rho > 0` to reach agreement across scenarios. Those variables fall back to the positive default — but we now say so out loud rather than silently.

## Changes

**Enforcement — hard error on `rho <= 0`:**
- Validate inputs: `defaultPHrho` (`attach_Ws_and_prox`; `None` is allowed when a `rho_setter` supplies every nonant — caught later if not), `--default-rho` (`decomp._get_rho_setter`), and each `rho_setter` output (`_use_rho_setter`).
- Central guard `check_rhos_positive()` runs **after Iter0 rho setup** and **before every solve loop** (once per PH iteration), so any rho-updating extension that drives rho `<= 0` is caught before it reaches the solver.

**Stop silently inflating / respect small computed rhos:** a shared rule in `utils/rho_utils.py` — `RHO_ZERO_TOL = 1e-8` (a numerical-zero tolerance, **not** a rho floor) and `resolve_rho(val, default)`:
- a computed rho **above** the tolerance is kept **as-is** — small but genuine rhos are no longer inflated up to the default (this was most visible in `sensi_rho`, which floored everything below `defaultPHrho`);
- a computed rho **at/below** the tolerance (≈0) or **negative** is uninformative, so it falls back to the positive default and is **reported** (rank-0, aggregated to a count, re-emitted only when the count changes — no log spam).

Applied uniformly to `coeff_rho`, `sep_rho`, `grad_rho`, and `sensi_rho`. For the sensi family the fallback is applied to the final value **after** the cross-scenario max, so a single near-zero scenario no longer inflates the max with the default. `grad_rho`'s negative results (sign cancellation when `grad_order_stat > 0.5`) now fall back gracefully instead of reaching the guard.

> **Re `reduced_costs_rho`:** this PR does not edit `reduced_costs_rho.py`; `ReducedCostsRho` only *inherits* the `_SensiRhoBase` change. It is being deprecated/removed in #759 (after which it errors in setup and never reaches this code). The two PRs touch disjoint files — no conflict, either merge order is fine.

## Tests

New `mpisppy/tests/test_rho_enforcement.py` — the guard, the `resolve_rho` rule (incl. "small positive kept, ≈0/negative defaulted"), a sensi-rho test driving the real cross-scenario max, the no-spam reporter, and `--default-rho` validation; plus solver-gated farmer integration tests. Wired into `run_coverage.bash` and `.github/workflows/test_pr_and_main.yml`. `test_sep_rho.py`'s zero-cost case now asserts the "uses default rho" contract; the `test_grad_rho_bundles.py` stand-in PH was extended for the option surface `_compute_and_update_rho` now reads.

All rho-related tests pass; ruff clean; no regressions in `test_ef_ph`, aircond `general_rho_setter`, APH, farmer CI, or `generic_cylinders`.


bknueven (approved):
Is there an opportunity to factor the similar logic out of the four rho setters? Not sure.

DLWoodruff:
Good call — done in 359232b9. Factored the shared tail (loop over nonants → `resolve_rho` → accumulate defaulted → `report_zero_rho_fallback`) into `rho_utils.assign_rho_with_fallback(opt, default_rho, source, report_state, value_fn, reason=...)`. Each of CoeffRho/SepRho/SensiRho/GradRho now just passes a small `value_fn(s, ndn_i)` for its own raw rho; behavior is unchanged (rho-enforcement + sep/grad/sensi tests pass, plus a CoeffRho end-to-end farmer run). Also merged current main to clear the conflict.

Issue #762, opened 2026-06-15
Adding compatibility checks for HiGHS (and perhaps other solvers?)


tvalenciaz:
Writing and running some examples with HiGHS, I identified a couple of issues where we could check compatibility from the start. I'm not sure if there was already a discussion about whether it is our job or the user's to verify that the solver they intend to use does support the features requested, but we could issue more helpful errors:

  1. If using HiGHS with PH and problem has integers/binaries. HiGHS does not support MIQP and just fails with `status= unknown, TerminationCondition=unknown`. It's true that if the user runs with `--tee`, then the solver does print "cannot solve MIQP with HiGHS". But we could make it easier for the user requiring `--linearize-proximal-terms` , or at least issue a more descriptive error. 
  2. HiGHS does not return reduced costs*, so `--reduced-cost-spoke` and related features cannot be used. 
  *I am not 100% sure about this, but I have noticed that the suffix attribute that is supposed to hold the reduced costs is empty when I use HiGHS and the error thrown is a bit cryptic. My hypothesis is that the solver does not return them, so we could issue a more helpful error saying that. This was tested with the hydro example in `generic_tester.py` as well as with the "relief" example introduced in #761

bknueven:
For (1), I think we should print a message if the solver fails at PH iteration 1 and we're using quadratic proximal terms. Something like "solver failed at PH iteration 1, likely because solver {solver} does not support quadratic objectives". Maybe if we got fancier we could try to extract the error message and re-raise it -- not sure where that message is coming from in the case of HiGHS.

For (2), this sounds like something that should be investigated and fixed in Pyomo. HiGHS surely reports reduced costs, Pyomo just isn't extracting them on request. Or, this could be an issue with mpi-sppy and the new Pyomo solver interfaces. I would suggest we investigate the root cause before putting in a HiGHS-specific patch.

DLWoodruff:
Investigated part 2 (reduced costs with HiGHS). The root cause is **not** HiGHS — HiGHS does report reduced costs. It's a bug in Pyomo's APPSI **legacy** solver wrapper that surfaces because of *how* mpi-sppy loads solutions, and it affects every APPSI solver (`appsi_gurobi` fails identically).

### Reproduction (LP with an `rc` IMPORT suffix)

| Path | primals | duals | reduced costs |
|---|---|---|---|
| `appsi_highs.solve(m)` (default `load_solutions=True`) | ok | ok | ok (`rc[x]=-1.0`) |
[...]

PR #768, opened 2026-06-18
Helpful error when a solver can't handle the PH quadratic proximal term (issue #762, part 1)

DLWoodruff:
## Part 1 of issue #762

When Progressive Hedging uses a quadratic proximal term and the solver cannot handle a quadratic objective, the run fails confusingly. For HiGHS on a problem with integers/binaries (an MIQP), the failure is the cryptic `status=unknown, TerminationCondition=unknown`. This adds an actionable message.

## Approach

PH's proximal term makes the subproblem objective quadratic unless `--linearize-proximal-terms` is set, so this is a quadratic-objective/solver mismatch. Pyomo's capability information is uneven (the legacy `has_capability` API reports it for solvers like cbc/glpk, but the APPSI and `pyomo.contrib.solver` wrappers used for HiGHS don't expose it), so a single up-front check isn't enough. Two complementary checks in `phbase.iterk_loop` cover both cases:

- **Proactive** — before the first proximal solve, if the solver *reports* (via `has_capability`) that it cannot handle a quadratic objective, raise immediately with guidance. Catches e.g. `cbc`, `glpk` before wasting a solve.
- **Reactive** — after the first proximal solve, if *no subproblem anywhere* produced a solution, raise with guidance. Covers solvers that expose no capability info (HiGHS et al.). The "no solution anywhere" test uses `allreduce_or`, so the raise decision is identical on every rank (no MPI desync).

Both checks no-op unless a non-linearized proximal term is actually active, and the reactive check is limited to iteration 1 — if the first quadratic solve succeeds, the solver supports quadratic objectives, so a later failure is a genuine optimization issue, not a capability problem. The error message points the user at `--linearize-proximal-terms` (and `--linearize-binary-proximal-terms` for binaries).

## Depends on #776 (merged)

These checks live at iteration 1, where the quadratic prox first becomes *active*. Previously the prox term was spliced into the objective in `PH_Prep`, so a `0 * x**2` term was structurally present already at iteration 0 — and cbc/glpk's LP writer raised `InvalidExpressionError` at `Iter0()` before the iteration-1 checks could run (reported in review below). #776 (*Defer attaching W and prox terms until after PH iteration 0*) makes iteration 0 structurally identical to the user's original objective, so iteration 0 is purely linear and the quadratic prox first appears at iteration 1 — exactly where these checks fire. This branch is rebased onto `main` with #776 included.

## Changes
- `mpisppy/utils/sputils.py`: `solver_quadratic_objective_capability()` — tri-state probe (`True`/`False`/`None`, where `None` means "unknown", never "unsupported").
- `mpisppy/phbase.py`: `_prox_is_quadratic`, `_check_prox_solver_capability` (proactive), `_check_prox_solve_succeeded` (reactive), `_reraise_as_prox_capability_error` (LP-writer raise → actionable message); wired into `iterk_loop`.
- `mpisppy/tests/test_prox_solver_compat.py`: solver-independent unit tests (probe tri-state, both guards, MPI-consistency via `allreduce_or`), wired into `run_coverage.bash` and `.github/workflows/test_pr_and_main.yml`.

## Validation
- New unit tests pass.
- Existing PH tests (`test_ef_ph`, `test_ph_main`, `test_prox_approx{,_e2e}`) pass; no false positives on a quadratic-capable solver (gurobi).
- End-to-end with **cbc** on farmer (post-rebase): without `--linearize-proximal-terms`, iteration 0 completes and PH then raises `Solver 'cbc' reports that it cannot handle a quadratic objective ... Re-run with --linearize-proximal-terms ...`; with `--linearize-proximal-terms`, the run completes.

### DIFF
--- a/.github/workflows/test_pr_and_main.yml
+++ b/.github/workflows/test_pr_and_main.yml
@@ -120,6 +120,10 @@ jobs:
         run: |
           coverage run $COV_ARGS -m pytest mpisppy/tests/test_rho_enforcement.py -v
 
+      - name: Test prox/solver quadratic compatibility checks
+        run: |
+          coverage run $COV_ARGS -m pytest mpisppy/tests/test_prox_solver_compat.py -v
+
       - name: Test generic_cylinders usage
         run: |
           coverage run $COV_ARGS -m pytest mpisppy/tests/test_generic_cylinders.py -v
--- a/mpisppy/phbase.py
+++ b/mpisppy/phbase.py
@@ -30,6 +30,14 @@ def profile(filename=None, comm=MPI.COMM_WORLD):
 logger = logging.getLogger('PHBase')
 logger.setLevel(logging.WARN)
 
+# Shared remediation hint for the quadratic-prox / solver compatibility checks
+# (issue #762): a solver that cannot handle the quadratic proximal term should
+# linearize it instead.
+_LINEARIZE_PROX_HINT = (
+    "Re-run with --linearize-proximal-terms (add "
+    "--linearize-binary-proximal-terms for binary variables)."
+)
+
 #======================
 
 def _Compute_Xbar(opt, verbose=False):
@@ -770,9 +778,7 @@ class PHBase(mpisppy.spopt.SPOpt):
             raise RuntimeError(
                 f"Solver '{self.options.get('solver_name')}' reports that it "
                 "cannot handle a quadratic objective, which the Progressive "
-                "Hedging proximal term requires. Re-run with "
-                "--linearize-proximal-terms (add "
-                "--linearize-binary-proximal-terms for binary variables)."
+                "Hedging proximal term requires. " + _LINEARIZE_PROX_HINT
             )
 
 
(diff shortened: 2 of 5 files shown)


DLWoodruff:
@tvalenciaz this addresses part 1 of #762 — please review it when you get a chance.

tvalenciaz:
This looks good to me! I tried it on the instances that prompted the issue submission and it works well.
However, for glpk and cbc, I tried running it and it seems to fail before it gets to the iteration 1 check. I am getting the message below from both cbc and glpk when running farmer (using the `farmer_generic.bash` script in the examples directory). 

I think during `Iter0()`, the quadratic term is turned off by being multiplied by 0, but it's still there, which both solvers complain about. But this message is not as cryptic as the other one, so maybe acceptable?

```
pyomo.common.errors.InvalidExpressionError: Model objective (Total_Cost_Objective) contains nonlinear terms that cannot be written to LP format
```

bknueven:
> I think during `Iter0()`, the quadratic term is turned off by being multiplied by 0, but it's still there, which both solvers complain about.

@DLWoodruff I thought we attach W and prox after iteration 0? If not, maybe we should just do that so the problem we solve at iteration 0 is effectively what the user passes us. This would be good for debugging, generally.

DLWoodruff:
The PR now addresses Tomas's issues with cbc and glpk (although we wonder if the latest Pyomo was running).
Ben's suggestion about when to attach W needs discussion (BTW: Pyomo removes terms that it can evaluation to zero so quadratic terms were not making it through to the solver on iteration 0; it was iteration 1, but I think that is now fixed with a proper error message)

DLWoodruff:
Converting to draft. With current Pyomo, the LP writer rejects the zero-coefficient quadratic prox term (`prox_on * x**2` with `prox_on=0`) that PH attaches before Iter0, so cbc/glpk now fail at **Iter0** — before any of this PR's `iterk_loop` guards can fire. The clean fix is structural and belongs with #772 (attach W and prox terms after iteration 0). Holding this PR until #772 lands, then revisiting whether these friendly-error guards are still needed and where they should live.

DLWoodruff:
@tvalenciaz thanks — and good catch on the cbc/glpk behavior. That was exactly the `0 * x**2` term still sitting in the iteration-0 objective: cbc/glpk's LP writer raised `InvalidExpressionError` at `Iter0()` before this PR's iteration-1 check could run.

That path is now closed upstream by #776 (*Defer attaching W and prox terms until after PH iteration 0*), merged today. With #776, iteration 0 is structurally the user's original objective — no quadratic term — so cbc/glpk solve iteration 0 cleanly and the quadratic prox first appears at iteration 1, exactly where this PR's checks live.

I've rebased this branch onto current `main` (which includes #776) and re-verified end-to-end with **cbc** on farmer:

- **without** `--linearize-proximal-terms`: iteration 0 now completes, then we get the actionable message
  `Solver 'cbc' reports that it cannot handle a quadratic objective ... Re-run with --linearize-proximal-terms (add --linearize-binary-proximal-terms for binary variables).`
- **with** `--linearize-proximal-terms`: completes normally (obj ≈ -112289.9).

So the cryptic `InvalidExpressionError` at `Iter0()` you hit is gone; cbc/glpk now reach the friendly message. Marking ready for review.

PR #769, opened 2026-06-19
Maximization: verify/guarantee max across all features (or a clear error)

DLWoodruff:
## Summary

Audits and hardens maximization (`sense=pyo.maximize`) support across mpi-sppy so that every feature either **works correctly** for a maximization problem or **raises a clear error** — never silently produces a wrong answer. ("Runs without error" is not the bar; correctness — e.g. bounds bracketing the optimum on the correct, sense-dependent side — is.)

The single source of truth for sense is `SPBase.is_minimizing` (mixed-sense scenarios already raise). The core turned out to be far more sense-aware than a "min-biased" reading suggests; the real risks were untested paths (now tested) and two clusters of genuinely-silent wrongness (now fixed or guarded).

Five logical commits, preserved for phase-by-phase review:

- **Phase 1 — tests for the correct-but-untested core.** End-to-end maximize tests for EF, PH, subgradient, L-shaped, FWPH (`mpisppy/tests/test_maximization.py`) plus a cylinders Lagrangian-bound test, each checking `max == -min` and that bounds bracket the optimum on the correct (sense-dependent) side. Fixes a stale "assumes minimization" comment in `opt/lshaped.py` (it negates the objective internally and supports max).

- **Phase 2 — confidence-interval silent-wrongness (bug fix).** The MMW / sequential-sampling gap correction was computed as if minimizing for *every* problem, due to three compounding bugs: a config-key typo (`pyo_opt_sense` read vs `pyomo_opt_sense` written), `is_minimizing` used without calling it, and the sense never being populated by the pipeline. `gap_estimators` / `mmw_ci` now pass and use the real sense (gap sign + optimality-gap *magnitude*), the wrong-sign warnings are rank-gated, and sequential sampling raises a clear error on maximization (its BM/BPL stopping rules assume a non-negative, shrinking gap).

- **Phase 3 — Pyomo agnostic guest.** Reads the model's real objective sense instead of hardcoding minimize; the AMPL guest's bare assert becomes a clear min-only `RuntimeError`.

- **Phase 4 — CVaR maximization (now subsumed by #751).** This phase independently added CVaR maximization via the mirrored lower-(reward-)tail Rockafellar–Uryasev formulation, but #751 (CVaR Phase 2/3) landed equivalent support on `main` first: the two are mathematically identical (#751 uses a `NonPositiveReals` excess with `δ_s ≤ Cost_s − η` and a unified `+β/(1-α)·δ_s` objective term; this phase used `NonNegativeReals` with `δ_s ≥ η − Reward_s` and `−β/(1-α)·δ_s`). On merging `main`, the `cvar.py` / `cvar_design.md` / `test_cvar.py` conflicts were resolved to `main`'s version (a strict superset of the tests here), so this phase contributes no net change — it is retained only as part of the audit's per-feature CVaR-max verification.

- **Phase 5 — doc + guest guards.** Adds `doc/designs/maximization_support.md` (per-feature support matrix). The gurobipy agnostic guest now raises a clear min-only error (it previously could silently mis-sign a maximization model); the GAMS guest already negates a `maximizing` model to a minimize internally, so it runs, but its bound sign-correctness under max is covered only by a smoke test and it is not maintained as a max surface.

## Result

- **Works + tested for max:** EF, PH, subgradient, L-shaped, FWPH, cylinder bounds, the MMW/CI gap pipeline, CVaR (via #751), the Pyomo agnostic guest. Works by construction: bundling, ADMM, chance constraints, linearized prox, mps_reader, the generic driver.
- **Fails loudly on max (intentional):** sequential sampling; the AMPL and gurobipy agnostic guests (guests are a minimize-only surface; the Pyomo guest is the exception, and the GAMS guest merely *runs* max by internal negation — smoke-tested, not a maintained max surface).

## Notes for the reviewer

- New serial tests are wired into `run_coverage.bash` and the `regression` job of `test_pr_and_main.yml`; cylinder/CI/CVaR/agnostic tests extend already-CI-wired files.
- The AMPL guard is exercised by the CI agnostic job (that backend isn't installed in the dev env used here); everything else was verified locally.
- The pre-existing `test_agnostic_gurobipy_PH` failure (solver-fragile, already marked "Test Case is failing") is untouched by this work.


mpisppy/confidence_intervals/seqsampling.py
```diff
@@ -185,6 +185,28 @@ def __init__(self,
...
+        # (multistage) models cannot be built from cfg alone at this point, so
+        # if that build fails we skip the check and let minimization (the common
+        # case) proceed.
+        try:
+            _sn = self.refmodel.scenario_names_creator(1)
+            _probe = self.refmodel.scenario_creator(
+                _sn[0], **self.refmodel.kw_creator(cfg))
+            _is_min = sputils.find_active_objective(_probe).is_minimizing()
+        except Exception:
+            _is_min = True
```
bknueven: This is expensive to just check if we're minimizing. I'm also not a huge fan of the fall-through, which just assumes minimization.

bknueven (commented):
I'm happy modulo the one comment.

mpisppy/confidence_intervals/seqsampling.py
DLWoodruff: Good points — both addressed (just pushed).

The eager probe (and its `except: _is_min = True` fall-through) is gone. The sense is now read from the first gap estimation, which already builds a real evaluator and knows the sense (`gap_estimators` returns `is_minimizing`; the multistage path exposes `ev.is_minimizing`):

- **Not expensive:** the common minimize path builds no extra model — the check piggybacks on a solve that happens anyway, and fires before the stopping loop ever uses `Gk`/`sk`.
- **No assume-minimization fall-through:** the sense comes from a genuinely-built model via a small `_min_only_guard` helper called from both `run()` paths; there's no swallow-all default that could let a max model slip through.

`test_seqsampling_maximize_raises` now drives a real maximize `run()` and asserts the guard fires on iteration 1.

PR #773, opened 2026-06-27
Consistent CROSS_SCENARIO_COST/CUT buffer sizing above one rank per cylinder (#727)

### LINKED ISSUE #727: CROSS_SCENARIO_COST buffer length is inconsistent (only correct at one rank per cylinder)
## Summary

The `Field.CROSS_SCENARIO_COST` buffer is sized three different ways, which
only agree when each cylinder has exactly **one rank**. With more than one
rank per cylinder the sizes diverge and cross-scenario cut runs break. This is
a pre-existing issue, independent of (but surfaced by) the flexible-rank work.

Let `nscen = len(all_scenario_names)` and `local = len(local_scenario_names)`.
The field is logically *per local scenario, one recourse-cost (eta) value per
global scenario*, i.e. a **local-sized** field of length `nscen * local`.

But the three sites disagree:

| Site | Length used |
|---|---|
| Sender — `extensions/cross_scen_extension.py:241-244` (`register_send_field`) | `nscen * nscen` |
| Receiver buffer — `cylinders/cross_scen_spoke.py:61` (`register_recv_field`, default length) | `_field_lengths[CROSS_SCENARIO_COST]` = `local * local` |
| Receiver read loop — `cylinders/cross_scen_spoke.py:167-172` | `nscen * local` |

The receiver's default comes from `cylinders/spwindow.py:62`:
```python
Field.CROSS_SCENARIO_COST : _total_number_scenarios * _total_number_scenarios
```
where `_total_number_scenarios` is set to **`len(opt.local_scenarios)`**
(`spwindow.py`, in `FieldLengths.__init__`) — i.e. the *local* count, despite
the "total" name. So the default is `local²`, not the intended `nscen * local`.

## Consequence

At one rank per cylinder, `local == nscen`, so all three collapse to `nscen²`
and everything works — which is presumably why this has gone unnoticed. With
more than one rank per cylinder (`local < nscen`):

- the sender advertises `nscen²` while the receiver expects `local²`, so
  `_validate_recv_field` should reject the mismatch at window creation; and
- the read loop assumes `nscen * local`, matching neither buffer.

So cross-scenario cuts are effectively limited to one rank per cylinder today.

## Fix direction

Size `CROSS_SCENARIO_COST` consistently as the local-sized `nscen * local`
on both the sender (`register_send_field` with `nscen * local_scen_count`) and
the receiver (explicit `nscen * local`), and correct/rename the mislabeled
`_total_number_scenarios` (it currently holds the local scenario count). This
also unblocks cross-scenario cuts under unequal per-cylinder rank counts
(flexible ranks).

DLWoodruff:
Fixes #727.

## Problem

`Field.CROSS_SCENARIO_COST` and `Field.CROSS_SCENARIO_CUT` were sized three
different ways that only agree when each cylinder has exactly **one rank**
(`local == nscen`). The root cause is `FieldLengths.__init__` in `spwindow.py`
setting `_total_number_scenarios` to the **local** scenario count
(`len(opt.local_scenarios)`) despite the "total" name.

With more than one rank per cylinder (`local < nscen`):

- **`CROSS_SCENARIO_COST`** (logically `nscen * local`): the sender advertised
  `nscen * nscen`, the receiver default was `local * local`, and the read loop
  assumed `nscen * local` — three different sizes. `_validate_recv_field`
  rejects the sender/receiver mismatch at window creation.
- **`CROSS_SCENARIO_CUT`** (logically `nscen * (nonants + 2)`): sender and
  receiver both took the default `local * (nonants + 2)`, but the spoke writes
  `nscen` rows — an out-of-bounds write.

## Fix

1. **`spwindow.py`** — `_total_number_scenarios` now holds the global count
   (`len(opt.all_scenario_names)`); the name is finally truthful. This corrects
   `CROSS_SCENARIO_CUT` to its global size. `CROSS_SCENARIO_COST` becomes
   `_total_number_scenarios * _local_scenario_length` (`nscen * local`), not
   squared.
2. **`cross_scen_extension.py`** — the sender registers `CROSS_SCENARIO_COST`
   as `nscen * local_scen_count` to match.

`_total_number_scenarios` is used only by these two fields, so the change is
safe for all other cylinder configurations.

## Test

`mpisppy/tests/test_cross_scenario_buffer_sizing.py` (serial, deterministic, 3
tests) drives the length computation and the extension's `register_send_fields`
with a stand-in `opt` where `local < nscen`, pinning all three sizings at
`nscen * local` / `nscen * (nonants + 2)`. It fails on the pre-fix code and
passes on the fix. Wired into `run_coverage.bash` and `test_pr_and_main.yml`.

## Note (out of scope)

A serial unit test was chosen deliberately. An end-to-end MPI run with more
than one rank per cylinder currently aborts **before** the buffer logic, at
`MPI.Win.Allocate` (`MPI_ERR_WIN`), and this reproduces on `main` without these
changes — a separate, pre-existing issue independent of #727 (the buffer fix is
necessary but not sufficient to fully exercise multirank cross-scenario). Cross
scenario at one rank per cylinder is unaffected and continues to work.

### DIFF
--- a/.github/workflows/test_pr_and_main.yml
+++ b/.github/workflows/test_pr_and_main.yml
@@ -98,6 +98,10 @@ jobs:
         run: |
           coverage run $COV_ARGS -m pytest mpisppy/tests/test_chance_constraint.py -v
 
+      - name: Test cross-scenario buffer sizing
+        run: |
+          coverage run $COV_ARGS -m pytest mpisppy/tests/test_cross_scenario_buffer_sizing.py -v
+
       - name: Test afew
         run: |
           PYARGS="-m coverage run $COV_ARGS"
--- a/mpisppy/cylinders/spwindow.py
+++ b/mpisppy/cylinders/spwindow.py
@@ -58,8 +58,13 @@ _field_lengths = {
         Field.OBJECTIVE_OUTER_BOUND : 1,
         Field.EXPECTED_REDUCED_COST : field_length_components._total_number_nonants,
         Field.SCENARIO_REDUCED_COST : field_length_components._local_nonant_length,
+        # CROSS_SCENARIO_CUT is global-sized: every rank holds one cut row (eta
+        # coef + nonant coefs + constant) for every scenario, replicated across a
+        # cylinder's ranks. CROSS_SCENARIO_COST is local-sized: one recourse-cost
+        # (eta) value per global scenario, for each *local* scenario. Both must use
+        # the global scenario count (_total_number_scenarios), not the local one.
         Field.CROSS_SCENARIO_CUT : field_length_components._total_number_scenarios * (field_length_components._total_number_nonants + 1 + 1),
-        Field.CROSS_SCENARIO_COST : field_length_components._total_number_scenarios * field_length_components._total_number_scenarios,
+        Field.CROSS_SCENARIO_COST : field_length_components._total_number_scenarios * field_length_components._local_scenario_length,
         Field.NONANT_LOWER_BOUNDS : field_length_components._total_number_nonants,
         Field.NONANT_UPPER_BOUNDS : field_length_components._total_number_nonants,
         Field.BEST_XHAT : field_length_components._local_nonant_length + field_length_components._local_scenario_length,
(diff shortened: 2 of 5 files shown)


DLWoodruff:
The pre-existing multirank `MPI.Win.Allocate` crash mentioned in the description (out of scope for this PR) is now tracked as #774.

bknueven (approved):
We should re-write cross-scenario cuts, and along the way L-Shaped:

1. For optimality cuts especially, cut aggregation makes more sense than tracking individual scenario recourse costs -- the later is not scalable.

2. While blended optimality / feasibility cuts make the algorithm simpler, some work since I implemented this suggests it's not as efficient as doing things right separately.

3. Feasibility cuts should be re-implemented to ensure they're strong and can generate facets or near-facets.

Basically, we should read this paper and apply it to mpi-sppy: https://www.research.unipd.it/bitstream/11577/3350084/1/CPXbenders.pdf

@tvalenciaz is this along the lines of what you were thinking?

PR #778, opened 2026-06-27
W-oscillation detection (PR1: detect + report)

DLWoodruff:
## Summary

First of two PRs adding a hub extension to **detect** (this PR) and later
**interrupt** oscillation / cycling in the PH dual weight (W) vector. This PR is
**pure observation** — it attaches no rho or fixing changes, so a run with
`--detect-W-oscillations` follows exactly the same optimization trajectory as
one without. With no flag, behavior is unchanged (backward compatible).

Design: `doc/designs/w_oscillation_design.md` (on the `oscillation` design branch).

## What's added

- New extension `mpisppy/extensions/w_oscillation.py` (`WOscillationMonitor`)
  with two JSON-selectable, pluggable detectors:
  - **`zero_crossings`** — a port of PySP's `sorgw` plugin: per (scenario,
    nonant) it counts sign changes of W and of the consecutive differences,
    plus a back-half/front-half `|ΔW|` damping ratio.
  - **`w_hash_recurrence`** — the Watson–Woodruff *Progressive Hedging
    Innovations* (Computational Management Science, 2011) §2.4 cycle detector:
    hash the per-scenario W vector for each nonant and flag a recurrence. In
    the distributed setting the signature is made independent of the
    scenario→rank map by summing identity-mixed per-scenario 64-bit hashes and
    reducing with `MPI.SUM` (an additive/multiset hash); `min_period ≥ 2` so
    convergence (constant W) is not mistaken for a cycle.
- Per-(scenario, nonant) verdicts are aggregated across scenarios with per-node
  `SUM`/`MAX` reductions; cylinder rank 0 writes the aggregate CSV (full
  variable name + evidence). An optional per-scenario detail CSV is gathered to
  rank 0.
- CLI: `--detect-W-oscillations <json>`, wired through `config.py` /
  `cfg_vanilla.py` / `generic/extensions.py`, mirroring the Slammer.
- Example control file `examples/sizes/config/w_oscillation.json`.

## Tests

`mpisppy/tests/test_w_oscillation.py` (wired into `run_coverage.bash` **and**
`test_pr_and_main.yml`): unit tests for the pure detectors (sign counts,
signature determinism + distribution-independence, recurrence vs. convergence,
config validation) and an end-to-end farmer run asserting the flag attaches the
extension and writes a well-formed CSV. Verified manually that the aggregate
report is byte-identical at `np=1` and `np=2`.

## Docs

A `w_oscillation` section in `extensions.rst` cross-referencing the
`wtracker_extension` (for the full W trajectory / moving statistics), plus flag
entries in the generic-cylinders option lists.

## Notes

- Scope is **synchronous PH**.
- This is the detection half; **interruption** (rho reduction / slamming via the
  Slammer action layer) is the planned follow-up PR.

### DIFF
--- a/doc/src/extensions.rst
+++ b/doc/src/extensions.rst
@@ -24,6 +24,8 @@ command-line flags:
 
 - ``--fixer`` -- activates the fixer extension
 - ``--slamming-directives-file <file>`` -- activates the slammer extension
+- ``--detect-W-oscillations <file>`` -- activates W-oscillation detection
+  (see :ref:`w_oscillation`)
 - ``--mipgaps-json <file>`` -- activates the mipgapper extension
 - ``--user-defined-extensions <module>`` -- loads a custom extension
 - ``--grad-rho`` -- activates gradient-based rho (see :ref:`rho_setting`)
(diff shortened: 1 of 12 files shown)


mpisppy/extensions/w_oscillation.py
```diff
@@ -0,0 +1,552 @@
...
+    diff_crossings = count_sign_changes(diffs, tol)
+
+    absdiffs = [abs(d) for d in diffs]
+    half = len(absdiffs) // 2
+    front = absdiffs[:half]
+    back = absdiffs[half:]
+    frontavg = sum(front) / len(front) if front else 0.0
+    diffs_ratio = (sum(back) / len(back) / frontavg) if (back and frontavg > tol) \
+        else 0.0
+
```
bknueven, line 146: I don't understand why we're splitting the `absdiffs` in half to do this calculation.

mpisppy/extensions/w_oscillation.py
```diff
@@ -0,0 +1,552 @@
...
+    # ------------------------------------------------------------------ #
+    def _reduce_per_node(self, nJ, arrays_and_ops):
+        """Allreduce each (array, op) over the per-node communicators, returning
+        rank-identical global arrays of length ``nJ``."""
+        outs = [np.zeros(nJ, a.dtype) for a, _ in arrays_and_ops]
+        for ndn, positions in self._node_pos.items():
+            comm = self.opt.comms[ndn]
+            pos = np.array(positions)
+            for (a, op), out in zip(arrays_and_ops, outs):
+                mpitype = MPI.INT if a.dtype == np.int32 else MPI.DOUBLE
```
bknueven: This looks weird -- I think its correct in context, but if this function gets re-used it could be a problem, or we run this code with a backwards incompatible version of `numpy` in 10 years.

Maybe we can build a map from `numpy` types to `MPI` types and throw an error if the lookup fails?

doc/src/extensions.rst
```diff
@@ -349,6 +351,100 @@ diagnostic tool intended for tuning rho and convergence behavior; it
...
+The detail file has columns ``iteration, node, scenario, variable, method,
+w_crossings, diff_crossings, diffs_ratio, w_value``.
+
+The aggregate CSV has a header row and one row per flagged nonant per
+detection event, with columns ``iteration, node, variable, method,
+num_scenarios_total, num_scenarios_flagged, max_w_crossings,
+max_diff_crossings, max_diffs_ratio, cycle_period``. The report is
+independent of how scenarios are distributed across MPI ranks.
+
+See ``doc/designs/w_oscillation_design.md`` for the full design.
```
bknueven, line 445: We should add this design document if we mention it.

mpisppy/extensions/w_oscillation.py
DLWoodruff: Good prompt — added a comment at the calculation site (1b915c45). `diffs_ratio` is a damping check: split `|ΔW|` into a front (older) and back (newer) half and compare their mean step sizes. If the swings are shrinking over time the back-half mean is small relative to the front-half mean (ratio well below 1); if they aren't damping out the ratio stays near or above 1. (Faithful port of sorgw's `Of_Interest` test.)

mpisppy/extensions/w_oscillation.py
DLWoodruff: Done in 1b915c45. Replaced the int32/else fallback with an explicit module-level `_NUMPY_TO_MPI` map ({int32: MPI.INT, float64: MPI.DOUBLE}); an unknown dtype now raises a `ValueError` naming the dtype instead of being silently treated as a double.

doc/src/extensions.rst
DLWoodruff: Added `doc/designs/w_oscillation_design.md` in 1b915c45 and refreshed its status header to reflect PR1 being implemented (PR2/interruption still to come).

bknueven (approved):

PR #779, opened 2026-06-29
Out-of-the-box auto-configuration

DLWoodruff:
## Out-of-the-box auto-configuration

Ready for review. This adds the `--out-of-the-box` (OOTB) mode described in
`doc/designs/out_of_the_box_design.md`. It lets a relatively new user obtain a
*defensible* mpi-sppy run with almost no knowledge of the library's internals:
supply a model module and scenario data, add `--out-of-the-box`, and the driver
introspects the environment + model and assembles a configuration (solver, EF vs.
decomposition, spokes, flexible rank split, proper bundling), prints the
**equivalent explicit command line**, runs it, and prints a **Suggestions** list.

**User-supplied options always win** — OOTB only fills gaps.

### What's in this PR

| Area | What |
|---|---|
| Interpreter + wiring | `mpisppy/generic/out_of_the_box.py` — pure `recommend(facts, policy) → Decision`, plus `gather_facts`, `apply_decision`, and the `configure`/`report_suggestions` driver hooks |
| CLI | `config.ootb_args()` — the three tier flags + `--inspect-only`, as optional-value flags (`str` + `nargs='?'`); wired through `generic/parsing.py` and `generic_cylinders.py` |
| Policy file | `mpisppy/generic/ootb_policies/ootb_policy_2026-06-28.json` — dated, declarative knowledge (solver order, EF budget, spoke ladder, rank ratios, effort model, extra-option defaults) |
| Validator | `mpisppy/generic/ootb_validate.py` — static schema + decision checks + an opt-in `--run` tier that flags problem runs (design §8) |
| Calibrator | `mpisppy/generic/ootb_calibrate.py` — fits the effort coefficients from timed example solves; its output *is* the shipped policy's numbers (design §9) |
| Docs | `doc/src/out_of_the_box.rst` (linking the design document), a tip in `generic_cylinders.rst`, and an `index.rst` entry. Deliberately **not** in the quick start: the feature is experimental and the page says so up front |

### Effort tiers

- `--out-of-the-box-minus` — instantiate nothing (structural decisions only; cannot bundle)
- `--out-of-the-box` (default) — one probe scenario → size-aware EF gate and bundling
- `--out-of-the-box-plus` — declared and accepted, but **behaves exactly like base**. The design (§5.2) is for a phase that is **not scheduled**; the flag exists so the vocabulary is stable, not because work is in flight

### Seconds-calibrated effort

The effort coefficients are produced by the calibration tool from timed EF solves
on the examples and kept in **seconds units**, so the EF budget reads as a
wall-clock target (`ef_effort_budget = 120`, i.e. ~120 s) rather than an opaque
number. The shipped numbers are this reference machine's (gurobi); they are
per-machine/solver and reproducible by re-running the calibrator. The resulting
gate is sensible: a continuous problem stays on the EF to a large scenario count,
but an integer-heavy MIP decomposes far sooner (the superlinearity is real).

### CI vs. local

The per-PR gate is solver-free and fast: the validator's **static** layer plus
the **synthetic-facts** decision layer, on every shipped policy, via
`mpisppy/tests/test_ootb_validate.py`; and the calibrator's **pure fit** via
`mpisppy/tests/test_ootb_calibrate.py`. Both are wired into `run_coverage.bash`
and `test_pr_and_main.yml`. Everything that needs a solver — example
instantiation, the validator's `--run` tier, and the calibrator's measurement —
runs locally / nightly, not in CI.

### Try it

```bash
# run OOTB on farmer (serial -> EF; mpiexec -> decompose when worth it)
python -m mpisppy.generic_cylinders --module-name farmer --num-scens 3 --out-of-the-box

# validate the shipped policy (fast, solver-free layers)
python -m mpisppy.generic.ootb_validate

# exercise the whole pipeline end to end on the examples (needs a solver)
python -m mpisppy.generic.ootb_validate --run --json report.json
```

### Since the first review

The branch has been through repeated review rounds. What they changed, beyond
the obvious polish:

- **Models that take no scenario-count flag now work.** `--num-scens` is
  declared by the *model*, not the driver, so netdes (two-stage, reads its count
  off `--instance-name`) used to abort. OOTB now asks `kw_creator` and, if the
  count is still unknown, says so instead of letting a model traceback escape.
- **The equivalent command line actually runs.** `--branching-factors` is a
  single `ListOf` token and was emitted unquoted, so pasting the line for any
  multistage model gave `unrecognized arguments: 3 2`; and the anchor emitted a
  derived `--num-scens` even for models that reject it. Both printed lines are
  now executed verbatim as a check.
- **The rank split models what the run does.** Spokes the user asks for that are
  not on the policy ladder are counted (they are cylinders too), the modelled
  order matches the launch order so the split names the right cylinders, and a
  ratio OOTB does not emit is not modelled.
- **An EF is no longer advertised under `mpiexec`.** When OOTB picks the
  extensive form the printed line is serial, and a run that does use several
  ranks for an EF says what that cost. See #877 for the general case.

Three defects found while testing this were **outside** OOTB and are filed
separately rather than fixed here: #875 (`--lagrangian` with `--relaxed-ph` or
`--ph-dual` aborts on duplicate `Field.DUALS` publishers), #876 (`--grad-rho`
with `--ph-xfeas-spoke` never registers `best_xhat_buf`), and #877. All three
reproduce on `main` with no OOTB option.

### Notes for reviewers

- Feedback especially welcome on the **policy schema** and the **spoke ladder /
  rank-ratio** cold-start choices, and on whether shipping a machine-calibrated
  policy in the repo is the right call (the alternative is to ship the tool and
  calibrate per environment).
- The tier flags and the policy schema are the parts hardest to change later;
  the effort numbers are re-derivable by re-running the calibrator.

### DIFF
--- a/.github/workflows/test_pr_and_main.yml
+++ b/.github/workflows/test_pr_and_main.yml
@@ -124,6 +124,10 @@ jobs:
         run: |
           coverage run $COV_ARGS -m pytest mpisppy/tests/test_generic_cylinders.py -v
 
+      - name: Test out-of-the-box interpreter wiring
+        run: |
+          coverage run $COV_ARGS -m pytest mpisppy/tests/test_out_of_the_box.py -v
+
       - name: Test out-of-the-box policy validator (layers 1 + 2-synthetic)
         run: |
           coverage run $COV_ARGS -m pytest mpisppy/tests/test_ootb_validate.py -v
--- a/.github/workflows/test_pr_and_main.yml
+++ b/.github/workflows/test_pr_and_main.yml
@@ -124,6 +124,10 @@ jobs:
         run: |
           coverage run $COV_ARGS -m pytest mpisppy/tests/test_generic_cylinders.py -v
 
+      - name: Test out-of-the-box policy validator (layers 1 + 2-synthetic)
+        run: |
+          coverage run $COV_ARGS -m pytest mpisppy/tests/test_ootb_validate.py -v
+
       - name: Test xhat from file
         run: |
           coverage run $COV_ARGS -m pytest mpisppy/tests/test_xhat_from_file.py -v
--- a/.github/workflows/test_pr_and_main.yml
+++ b/.github/workflows/test_pr_and_main.yml
@@ -128,6 +128,10 @@ jobs:
         run: |
           coverage run $COV_ARGS -m pytest mpisppy/tests/test_ootb_validate.py -v
 
+      - name: Test out-of-the-box effort calibrator (pure fit)
+        run: |
+          coverage run $COV_ARGS -m pytest mpisppy/tests/test_ootb_calibrate.py -v
+
       - name: Test xhat from file
         run: |
           coverage run $COV_ARGS -m pytest mpisppy/tests/test_xhat_from_file.py -v
--- a/doc/designs/out_of_the_box_design.md
+++ b/doc/designs/out_of_the_box_design.md
@@ -4,9 +4,10 @@
 the DLWoodruff fork. Proceeding deliberately ("slowly"): requirements confirmed
 and the major design questions resolved — decision-logic mechanism (§5), policy
 file + path selection (§5.1), effort tiers (§5.2), bundle sizing & the EF gate
-(§5.3), `--inspect-only` (§5.4), rank allocation (§5.5), and the PR1 validator
-(§8). The first dated policy file is committed. No production library code yet —
-only an uncommitted interpreter *sketch* (§7).
+(§5.3), `--inspect-only` (§5.4), rank allocation (§5.5), and the PR1 tooling —
+validator (§8) and effort calibrator (§9). The first dated policy file is
+committed. No production library code yet — only an uncommitted interpreter
+*sketch* (§7).
 **Author:** dlw (captured with Claude Code assistance)
 **Last updated:** 2026-06-28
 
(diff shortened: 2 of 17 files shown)


DLWoodruff:
Thanks — all seven reproduce, and all seven are fixed in 4d899f82. Two of them had survived several earlier review passes.

**Exclude HiGHS from MIQP and linearize PH prox for MIP.** Confirmed against mpi-sppy's own source (`phbase.py`: *"HiGHS, which cannot solve an MIQP"*), which the policy's `caveats.highs` already acknowledged while the MIQP list included it anyway. Dropped from the MIQP class list. For the MIP route I linearized rather than skipping the solver: a new `no_miqp_force_linearize_prox` (`appsi_highs`, `highs`) triggers `--linearize-proximal-terms` when the model has integers, so highs stays usable instead of being declared unusable — the same remedy already applied to cbc/glpk. `lp_mip_only_force_linearize_prox` keeps its narrower meaning: those two cannot take a quadratic prox at all, highs only when integers are present.

**Separate EF and decomposition solver overrides.** `gather_facts` read `cfg.get("solver_name") or cfg.get("EF_solver_name")`. Now tracked as two fields. `--EF-solver-name` is still honoured when the EF is certain — below the rank floor there is no decomposition to choose a solver for — and on the size-gated EF path as before.

**Validate policy block types before dereferencing.** Reproduced: `"solver": []` gave `AttributeError: 'list' object has no attribute 'get'`. The validator's job is to report, not crash. Block shapes are checked before anything dereferences them, and `spoke_ladder.rungs` is confirmed to be a list.

**Restrict ladder rungs to spoke flags.** The check tested `DECOMPOSITION_FLAGS` although its own name said "(spoke) flags". Now `SPOKE_FLAGS`.

**Recognize `--fwph-objgap-hub` as a decomposition hub flag.** Real in `config.py:956`, `generic/hub.py:49`, and config's own `HUBS` list. Added.

**Pass completed run outcomes to `report_suggestions`.** I took the second option you offered and removed the generator plus the claim, rather than building outcome capture. Neither call site passes an outcome, so `_sg_from_outcome` was dead code advertising a capability; wiring real convergence/iteration/gap data through the driver is more than this PR should carry, and no further phase is scheduled. The `outcome` parameter stays so adding it later is not a signature change, a test now asserts no generator reads it, and the page and the design document no longer say suggestions reflect how the run went.

**Correct overstated policy calibration claim.** Fixed. Only the `effort_scaling` coefficients are calibrated; the thresholds and budget magnitudes are authored and flagged in each block's `_cold_start_guess`, which is what the design document already said.

Each of the five behavioural fixes has a test that fails against the previous commit (`TestCopilotReviewFindings`). 131 tests pass, ruff clean, docs build, and the shipped policy validates at 103 checks.

PR #798, opened 2026-07-13
one sided test prints a message on success

bknueven:


mpi_one_sided_test.py
```diff
@@ -63,6 +63,8 @@ def main():
 
     del buff # Important!
     win.Free()
+    if (rank == 1):
+        print("Test passed. You might have an MPI installation that will work well.")
```
bknueven:
```suggestion
        print("Test passed. You might have an MPI installation that will work.")
```

DLWoodruff (approved):

PR #803, opened 2026-07-16
Add console-script entry points for the main drivers

DLWoodruff:
## What

Adds console-script entry points so users who `pip install` mpi-sppy (or `pip install -e .` from a clone) can run the main drivers as commands on their `PATH`, without having to locate the files in site-packages:

| Command | Equivalent to |
|---|---|
| `mpi-sppy-generic-cylinders` | `python -m mpisppy.generic_cylinders` |
| `mpi-sppy-mrp-generic` | `python -m mpisppy.mrp_generic` |
| `mpi-sppy-one-sided-test` | `python -m mpi4py mpi_one_sided_test.py` |

## Changes

- **`pyproject.toml`**: add `[project.scripts]` and an explicit `[tool.setuptools]` package/module declaration. The package list matches the previous `find_packages()` result exactly (verified by building a wheel) and additionally installs the top-level `mpi_one_sided_test.py` so its console script can import it.
- **`mpisppy/generic_cylinders.py`, `mpisppy/mrp_generic.py`**: move the `__main__` body into a `main()` callable so the entry points have a target. The `__main__` guard still calls `main()`, so the existing `runpy`-based `__main__` tests (`test_generic_cylinders.py`, `test_mrp_generic.py`) continue to exercise the refactored code.
- **Docs / README**: advertise the console scripts as the standard, location-independent invocation. Because `pip install -e .` also creates them, there is no pip-vs-source distinction to document.

## Notes

- For multi-rank **parallel** runs, the `python -m mpi4py` module form is still recommended (e.g. `mpiexec -np 3 python -m mpi4py -m mpisppy.generic_cylinders ...`) so that mpi4py installs its abort-on-exception handler. The docs keep that guidance.
- Command names are prefixed with `mpi-sppy-` to avoid putting generic names like `generic_cylinders` on users' `PATH`.

## Testing

- Built a wheel and confirmed all three scripts appear in `entry_points.txt` and that every `mpisppy` subpackage is still packaged.
- `pip install -e .` regenerates the three launchers on `PATH`; each runs.
- Docs build clean (`cd doc && make html`), with no warnings on the edited pages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,11 +35,11 @@ Here are two methods that seem to work well for installation, at least when cons
    ```
    Run the line aborve after cloning and moving to the repo root directory.
 
-To test your installation, cd to the directory where you installed mpi-sppy
-(it is called ``mpi-sppy``) and then give this command.
+To test your installation, give this command (once mpi-sppy is installed
+it works from any directory):
 
 ```
-mpirun -n 2 python -m mpi4py mpi_one_sided_test.py
+mpirun -n 2 mpi-sppy-one-sided-test
 ```
 
 If you don't see any error messages, you might have an MPI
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -55,6 +55,26 @@ Installing mpi-sppy
...
+puts a few console scripts on your `PATH`, so you can run the main drivers
+without locating the files in your environment:
+
+- `mpi-sppy-generic-cylinders` — the general-purpose driver, equivalent to
+  `python -m mpisppy.generic_cylinders` (see the `generic_cylinders` docs).
+- `mpi-sppy-mrp-generic` — the sequential-sampling (MRP) driver, equivalent
+  to `python -m mpisppy.mrp_generic`.
+- `mpi-sppy-one-sided-test` — the MPI one-sided diagnostic shown above.
+
+For multi-rank parallel runs, the `python -m mpi4py` module form is still
```
bknueven, line 72: It makes sense you can do this -- I wonder if we should update the documentation and examples to do so. I'd think this would be the preferred way of running so you don't need to even know where mpi-sppy is installed.

Or, I wonder if we could create our own script (a little harder than entry points) which effectively did `python -m mpi4py -m mpisppy.generic_cylinders` for the user ... beyond the scope of this PR, for sure.

bknueven (approved):

PR #810, opened 2026-07-21
Convert timed mipgap to use termination callback mechanics

bknueven:


DLWoodruff (commented):
config.checker needs two updates:
0. --timed-mipgap-options given without --timed-mipgap is an error (does nothing; alternatively, you could make it do something)
1. fwph_obj_gab_hub is a hub

bknueven:
> config.checker needs two updates: 0. --timed-mipgap-options given without --timed-mipgap is an error (does nothing; alternatively, you could make it do something)
> 
>     1. fwph_obj_gab_hub is a hub

Not sure about (1) (at least for this PR), but for (0) we went with just `--timed-mipgap` as the option.

mpisppy/utils/callbacks/termination/solver_callbacks.py
```diff
@@ -37,43 +41,95 @@ def __call__(self):
...
-    # checks if the time limit has been reached. So broader than what is presently
-    # needed for our present MIP-based use cases.
-    xpress_problem.addcbchecktime(cbchecktime_callback, user_termination_callback, 0)
+            xpress_problem.interrupt(xp.StopType.USER)
+        return None
+
+    # Per the Xpress documentation, this callback is invoked every time the
+    # Optimizer checks if the time limit has been reached. This is broader than
+    # what is presently needed for our MIP-based use cases.
+    xpress_problem.addCheckTimeCallback(
```
tvalenciaz: Using xpress 9.7.1 (the most recent my license allows), I had to change this to `xpress_problem.addcbchecktime(` for it to work. 
I know from the email thread that this was running for you so it's probably due to the older version. Do we want to check which of `addcbchecktime` or `addCheckTimeCallback` exists and call that one?

tvalenciaz (commented):
I only tested gurobi and xpress because I don't have CPLEX in my system. I had an issue with xpress, probably due to having to use an older version (9.7.1), but I think it would be easy to keep it compatible with both versions. Otherwise looks good to me!

mpisppy/utils/callbacks/termination/solver_callbacks.py
bknueven: Yes, we can check if `addCheckTimeCallback` exists, and if it does, call it, and if it doesn't call `addcbchecktime`.

tvalenciaz: Opened #832 doing this

Issue #828, opened 2026-08-11
Document that scenario models must be dill-serializable (a Pyomo rule closing over `cfg` silently breaks pickling)


DLWoodruff:
> **Edited 2026-08-11.** The original description claimed a Pyomo `ConfigDict`
> fails under `dill` but pickles fine with the stdlib. That was wrong (it came
> from serializing the same object twice in one script). The root cause is now
> understood and the description below has been rewritten; the correction is
> also recorded in [this comment](https://github.com/Pyomo/mpi-sppy/issues/828#issuecomment-5256969688).
> The "clearer error" half of this issue was resolved by #831. **What remains
> is the documentation.**

## Summary

Several mpi-sppy features serialize a scenario model with `dill`
(`utils/pickle_bundle.py::dill_pickle`, used by `generic/scenario_io.py` for
`--pickle-scenarios-dir`, `--pickle-bundles-dir`, and their `--unpickle-*`
counterparts; checkpoint/resume will use it too). There is a modeling pattern
that makes a scenario model **unserializable**, it is easy to write by
accident, and it is documented nowhere:

> A Pyomo rule defined as a **nested function that closes over the mpi-sppy
> `Config` object** (conventionally `cfg`).

Pyomo keeps the rule function on the component it built, so the closure pulls
the `Config` into the model's serialization graph — and a `Config` is a Pyomo
`ConfigDict`, which does not reliably survive serialization.

## Reproducer

```python
import io, dill, pyomo.environ as pyo
from mpisppy.utils.config import Config

def build():                          # nested, as inside a scenario_creator
    cfg = Config(); cfg.popular_args()
    m = pyo.ConcreteModel()
    def rule(m, i):                   # closes over cfg
        return (0, 10 if cfg.max_iterations else 5)
    m.v = pyo.Var([1, 2], bounds=rule)
    return m

dill.dump(build(), io.BytesIO())
# _pickle.PicklingError: args[0] from __newobj__ args has the wrong class
```

Hoist the value out of the closure and it round-trips fine:

```python
limit = cfg.max_iterations            # plain local
def rule(m, i):
    return (0, 10 if limit else 5)
```

The rule must be **nested** to reproduce. A rule at module scope captures a
global instead, and the failure does not occur.

## Root cause (upstream, in Pyomo)

Not a dill quirk — stdlib `pickle` fails identically. A freshly declared
`ConfigValue` is really an `UninitializedConfigValue`
[...]

PR #830, opened 2026-08-11
stoch_distr: hoist cfg out of the Pyomo rule closures so the models can be serialized

### LINKED ISSUE #829: stoch_distr example: hoist `cfg` out of the Pyomo rule closures so its models can be serialized
Split out of #828 as the concrete code fix, because it also gates work
elsewhere and deserves to be findable on its own.

## What

`examples/stoch_distr/stoch_distr.py::min_cost_distr_problem` defines
`slackBounds_rule` as a nested function that reads `cfg.ensure_xhat_feas`.
Closing over `cfg` pulls the mpi-sppy `Config` into the model's serialization
graph, and Pyomo's `ConfigDict` does not survive `dill` (background and
reproducer in #828). The result: a `stoch_distr` scenario model cannot be
dilled at all — with or without the ADMM wrapper, and independent of PH.

```python
model = stoch_distr.scenario_creator(
    "ADMM_STOCH__ADMM__Region1__ADMM__StochasticScenario1", **kw)
dill.dump(model, io.BytesIO())
# _pickle.PicklingError: args[0] from __newobj__ args has the wrong class
```

## Fix

Read the values the rules need into plain locals before defining them, so the
closures capture a `bool` rather than the `Config`:

```python
ensure_xhat_feas = cfg.ensure_xhat_feas      # hoisted
def slackBounds_rule(model, n):
    ...
    if ensure_xhat_feas:
        ...
```

Worth a sweep of the file for any other rule closing over `cfg`, and of the
other examples for the same shape.

## Why it matters beyond tidiness

1. It is the in-repo demonstration of the anti-pattern #828 asks to document —
   an example teaching the thing the docs warn against.
2. It blocks the checkpoint/resume rollout (#777). The design names
   `mpisppy/tests/examples/stoch_distr` as the vehicle for validating
   stoch-ADMM checkpointing in phase 2 — the wrapper-mutated model dill
   round-trip, variable-probability masks surviving restore, wrapper-held
   models released on resume. None of that can be tested while the underlying
   model cannot be serialized, so this fix is now a prerequisite of that phase
   rather than a step inside it.
3. Users hitting it through `--pickle-scenarios-dir` today get an opaque
   `PicklingError` pointing at neither the rule nor the config.

## Suggested acceptance

A test asserting that a `stoch_distr` scenario model round-trips through
`dill` — cheap, and it keeps the anti-pattern from creeping back.

DLWoodruff:
Fixes #829. Background and reproducer in #828.

## The problem

Both copies of the `stoch_distr` example defined Pyomo rules as nested
functions reading `cfg` directly. Pyomo keeps a reference to a rule function on
the model, so such a rule drags the whole `Config` into the model's
serialization graph — and a Pyomo `ConfigDict` **cannot be serialized with
dill**, though it pickles fine with the stdlib. A `stoch_distr` scenario model
therefore could not be dilled at all, with or without the ADMM wrapper:

```python
model = stoch_distr.scenario_creator(sname, **stoch_distr.kw_creator(cfg))
dill.dump(model, io.BytesIO())
# _pickle.PicklingError: args[0] from __newobj__ args has the wrong class
```

That silently breaks every feature built on dilling scenario models —
`--pickle-scenarios-dir` and `--pickle-bundles-dir` today (via
`utils/pickle_bundle.py`), and checkpoint/resume next (#777) — and the symptom
is an opaque `PicklingError` naming neither the config nor the rule.

## The fix

The rules read what they need into plain locals before being defined, so they
close over ints, floats and bools rather than over the `Config`:

```python
initial_seed = cfg.initial_seed      # hoisted out of the closure
spm = cfg.spm
cv = cfg.cv
...
def FlowBalance_rule(m, n):
    np.random.seed(scennum + node_num + initial_seed)
```

This is semantics-preserving: the same values are read from the same `cfg`
during the same `scenario_creator` call, only earlier in it. A `ConcreteModel`
evaluates its rules during construction either way, and nothing mutates `cfg`
in between.

**Both** copies were affected, not only the one named in the issue — found with
an AST sweep rather than by eye:

- `examples/stoch_distr/stoch_distr.py` — `slackBounds_rule`
  (`cfg.ensure_xhat_feas`) and `FlowBalance_rule` (`cfg.mnpr`,
  `cfg.initial_seed`, `cfg.spm`, `cfg.cv`)
- `mpisppy/tests/examples/stoch_distr/stoch_distr.py` — `FlowBalance_rule`
  (`cfg.initial_seed`, `cfg.spm`, `cfg.cv`)

## Tests

Two guards in `mpisppy/tests/test_stoch_admmWrapper.py` (already CI-wired, so
no new wiring):

1. `test_scenario_model_round_trips_through_dill` — dumps and reloads a
   scenario model.
2. `test_no_pyomo_rule_closes_over_cfg` — walks the module's AST for any nested
   function closing over `cfg`, catching the pattern structurally even if
   dill's behavior changes.

I reintroduced the anti-pattern to confirm guard 2 actually fails, and it does,
naming the offending rule. Worth noting: guard 1 alone passed during that check
because pytest had the fixed module cached in `sys.modules` while the AST guard
re-read from disk — in CI's fresh process both catch it, but the structural one
is the reliable guard.

## Verification

- Both copies now dill and reload (8,997 and 10,222 bytes); both raised
  `PicklingError` before.
- `test_stoch_admmWrapper.py`: 36 passed, 1 skipped.
- `test_admmWrapper.py`: 22 passed, 1 skipped.
- A live 5-iteration `--stoch-admm` run on the example reaches the same bound
  as before the change (`-71910.3395`).

`ruff check` clean.

### DIFF
--- a/examples/stoch_distr/stoch_distr.py
+++ b/examples/stoch_distr/stoch_distr.py
@@ -68,6 +68,19 @@ def min_cost_distr_problem(local_dict, cfg, stoch_scenario_name, sense=pyo.minim
     """
     scennum = sputils.extract_num(stoch_scenario_name)
 
+    # Read every cfg value the Pyomo rules below need into plain locals, so
+    # that the rules close over ints/floats/bools rather than over cfg itself.
+    # Pyomo keeps a reference to a rule function on the model, so a rule that
+    # closes over cfg drags the whole Config into the model's serialization
+    # graph -- and a Pyomo ConfigDict cannot be serialized with dill, which
+    # makes the scenario model unusable with --pickle-scenarios-dir and with
+    # checkpointing. See issue #829.
+    ensure_xhat_feas = cfg.ensure_xhat_feas
+    mnpr = cfg.mnpr
+    initial_seed = cfg.initial_seed
+    spm = cfg.spm
+    cv = cfg.cv
+
     # Assert sense == pyo.minimize, "sense should be equal to pyo.minimize"
     # First, make the special In, Out arc lists for each node
     arcsout = {n: list() for n in local_dict["nodes"]}
@@ -89,7 +102,7 @@ def min_cost_distr_problem(local_dict, cfg, stoch_scenario_name, sense=pyo.minim
         elif n in local_dict["buyer nodes"]:
             return (local_dict["supply"][n], 0)
         elif n in local_dict["distribution center nodes"]:
-            if cfg.ensure_xhat_feas:
+            if ensure_xhat_feas:
                 # Should be (0,0) but to avoid infeasibility we add a negative slack variable
                 return (None, 0)
             else:
@@ -108,7 +121,7 @@ def min_cost_distr_problem(local_dict, cfg, stoch_scenario_name, sense=pyo.minim
     model.FirstStageCost = pyo.Expression(expr=\
                     sum(local_dict["production costs"][n]*(local_dict["supply"][n]-model.y[n]) for n in local_dict["factory nodes"]))
     
-    if cfg.ensure_xhat_feas:
+    if ensure_xhat_feas:
         # too big penalty to allow the stack to be non-zero
         model.SecondStageCost = pyo.Expression(expr=\
                         sum(local_dict["flow costs"][a]*model.flow[a] for a in local_dict["arcs"]) \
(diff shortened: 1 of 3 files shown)


DLWoodruff:
**Correction to one sentence in this PR's description.**

It says a Pyomo `ConfigDict` "cannot be serialized with dill, though it pickles
fine with the stdlib". The second half is false — measured with a fresh object
each time, a populated `ConfigDict` fails under **both** `dill` and the stdlib
`pickle`. The earlier result came from testing both libraries against the same
object in one script, where the first attempt fails and the second succeeds
regardless of order. Details and the measurement table: #828 (comment).

Nothing in this PR's code or tests changes as a result — the models here still
could not be dilled, the hoisting fix is still correct and still
semantics-preserving, and the verification stands. Only the explanatory
sentence was wrong.

PR #846, opened 2026-08-19
PH_Prep refuses a second run

DLWoodruff:
## `PH_Prep` refuses a second run

Calling `ph_main()` twice on the same object silently solved a different
problem than the caller asked for.

`attach_Ws_and_prox` re-declared `W` and `rho` as fresh Params, discarding
whatever the first run put in them, and `attach_PH_to_objective` appended a
*second* PH term to the same `saved_objectives` entry, leaving the first live
and anchored to the previous run's `xbar`. Measured on farmer:

```
W after writing into it:  [8.374, 33.574, -41.948]
W after a second PH_Prep: [0.0,    0.0,     0.0]      (and a new Param id)
objective W_on count:     1 -> 2
```

No error, no warning, wrong answer. Split out of #799, where an in-place
re-solve was a prerequisite; that PR shipped EF-only.

`PH_Prep` now raises on a second call. That is the whole change — 26 lines in
`phbase.py`. A PH object solves one problem, and the W, rho and objective
terms it carries belong to that run.

### Why refuse rather than support it

I wrote the re-entrant version first: preserve the Params instead of
re-declaring them, track per scenario what each objective already carries,
refuse the option changes a restart cannot honor (`attach_prox`,
`options["smoothed"]`, `defaultPHrho`, `linearize_proximal_terms`), keep those
refusals consistent across ranks, survive an attach that raises partway
through the scenarios, and repair the ten or so extensions that keep per-run
state. It works, and it is about 800 lines in `phbase.py` and `spopt.py` plus
the extension sweep.

Nothing wants it. Every place in the tree that runs PH more than once builds a
new object — `examples/sizes/sizes_demo.py` does it six times — and
`APH_main` and `fwph_main` cannot be re-entered at all for reasons of their
own. So this PR buys the same protection against the silent wrong answer
without asking the project to carry machinery no caller exercises.

#848 records what re-entrancy would cost, extension by extension, so that work
does not have to be rediscovered if a caller ever turns up.

### Tests

`mpisppy/tests/test_ph_prep_once.py`:

- a second `PH_Prep` is refused
- the refusal leaves the first prep alone — the `W` Param is not replaced, the
  values written into it survive, and no second term appears in the objective
- a second `ph_main()` is refused, which is how a user would meet it
- one run still works, and its trivial bound is still a bound

Three of the four fail without the change.

`test_ph_prep_once` / `test_ef_ph` / `test_aph` / `test_outer_bound_only` /
`test_extensions` / `test_ph_extensions` / prox / smoothing / maximization /
config / proper_bundler / gradient_rho / generic_cylinders / grad_rho_bundles /
reduced_costs_fixer / rho_enforcement / solver_options_layers / slammer /
w_oscillation: 591 passed, 3 skipped. `test_with_cylinders` and the FWPH-spoke
cylinder tests pass on two ranks. `ruff` clean.

### Related

- #847 — four bugs found alongside this one that bite on a *single* run
- #848 — what supporting a second run would take

### DIFF
--- a/mpisppy/phbase.py
+++ b/mpisppy/phbase.py
@@ -1025,7 +1025,7 @@ class PHBase(mpisppy.spopt.SPOpt):
 
         Note:
             PH_Prep may be run only once on an object, so a PH object solves
-            one problem. Running it twice used to look like it worked and
+            one problem. Running PH_Prep twice used to look like it worked and
             quietly solve something else: attach_Ws_and_prox re-declared W and
             rho as fresh Params, discarding the duals the first run produced,
             and attach_PH_to_objective appended a *second* PH term to the same
@@ -1035,8 +1035,8 @@ class PHBase(mpisppy.spopt.SPOpt):
             W_on count from 1 to 2, with no error and no warning.
 
             Making it genuinely re-entrant is possible but is not free -- see
-            issue #848 -- and nothing in the tree wants it: every caller that
-            runs PH more than once builds a new object. So it refuses.
+            issue #848 -- and no caller wants it: every one that runs PH more
+            than once builds a new object.
         """
         if self._PH_prep_done:
             raise RuntimeError(
@@ -1067,11 +1067,8 @@ class PHBase(mpisppy.spopt.SPOpt):
         if not self._deferred_ph_attach:
             self.attach_PH_to_objective(attach_duals, attach_prox, attach_smooth)
 
-        # Last, so that a prep which died on the way here -- a rho_setter that
-        # raised, a model the prox approximation cannot take -- does not leave
-        # the object claiming a prep it never finished. Retrying then meets
-        # whatever actually went wrong instead of "already been run", which
-        # would point at the wrong fix.
+        # Last, so a prep that raised leaves the object unprepped and the
+        # retry reports the real failure, not "already been run".
         self._PH_prep_done = True
 
 
--- a/mpisppy/phbase.py
+++ b/mpisppy/phbase.py
@@ -301,6 +301,10 @@ class PHBase(mpisppy.spopt.SPOpt):
         self.convobject = None  # PH converger
         self.attach_xbars()
 
+        # Whether PH_Prep has run. It refuses a second time: this object's W,
+        # rho and objective terms belong to one run. See PH_Prep.
+        self._PH_prep_done = False
+
     @property
     def iter0_solver_options(self):
         """Read-only fold of solver_options_layers for iteration 0.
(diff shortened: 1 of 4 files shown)


mpisppy/phbase.py
```diff
@@ -1020,9 +1087,42 @@ def PH_Prep(
...
+            # re-enables both at the end.
+            #
+            # This makes a second ph_main() a restart, not a resume: Iter0
+            # re-solves the user's objective, the scenarios scatter, and
+            # iteration 1's Update_W adds rho*(x - xbar) on top of the W that
+            # was carried over. The carried W therefore shifts the starting
+            # duals but never enters a solve on its own. PH converges from any
+            # dual point, so this is correct either way; resuming instead
+            # would mean skipping Iter0, which ph_main is not structured for.
+            self.disable_W_and_prox()
```
bknueven: Why are we disabling W and prox here? It's unclear.

DLWoodruff: Fair — the comment explained what the call does to the objective without ever saying why it is there. The reason is that `PH_Prep` always hands `Iter0` a model with the dual and prox terms off. On a first prep that comes for free from `attach_Ws_and_prox`, which creates the Params at 0:

```python
# create ph objective terms, but disabled
scenario._mpisppy_model.W_on    = pyo.Param(initialize=0, mutable=True, within=pyo.Binary)
scenario._mpisppy_model.prox_on = pyo.Param(initialize=0, mutable=True, within=pyo.Binary)
```

A re-prep reuses those Params rather than re-declaring them (that reuse is the fix), and the previous run's `Iter0` left them at 1 via `reenable_W_and_prox`, so the 0 has to be written back. Same starting state either way, reached two ways.

Rewritten to say that in a52a50a1, and the restart-vs-resume paragraph the comment was carrying moved to the PR description.

I did consider dropping the `if reprepping` guard and calling `disable_W_and_prox()` unconditionally, since on a first prep it would write 0 over 0. The thing that stopped me is that `_disable_W`/`_disable_prox` fire `callout_agnostic` per scenario, so on a first prep that would now run before `attach_PH_to_objective`. It looks safe — the guests set `W_on`/`prox_on` in `attach_Ws_and_prox`/`create_ph_model` — but it changes callout order on the AMPL/GAMS/gurobipy paths, which this PR otherwise does not touch. Happy to do it if you would rather have the branch gone.

mpisppy/phbase.py
```diff
@@ -1020,9 +1087,42 @@ def PH_Prep(
...
+            # detaching. Refuse rather than silently run with the previous
+            # run's terms -- e.g. ph_main(attach_prox=False) after a plain
+            # ph_main() would otherwise still solve with the prox term, and a
+            # changed options["smoothed"] would be a no-op in both directions.
+            raise RuntimeError(
+                "PH_Prep was already run on this object with "
+                f"(attach_duals, attach_prox, attach_smooth)="
+                f"{self._PH_prep_request}; it cannot be re-run as {request}. "
+                "Build a new PH object for a different set of PH terms.")
+        self._PH_prep_request = request
```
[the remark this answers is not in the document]
DLWoodruff: Confirmed and fixed in ef252163. Reproduced with `PHIterLimit=5`, `time_limit=3600`, and the timer left where the first run put it:

```
[    0.21] Entering solve loop in PHBase.Iter0
[    0.22] Time limit 3600 seconds reached.
```

The second run solves `Iter0` and then breaks at iteration 1, never solving the augmented problem.

The reset went into `PH_Prep` under `reprepping` rather than into `ph_main`, since every entry point that can re-prep goes through it — `fwph_main`, the PH dual / xfeas / subgradient spokes, and any driver of the user's own — and all of them call `PH_Prep` exactly once per run, so `reprepping` is precisely "a new run is starting". It emits a `global_toc` when a time limit is in force, so the clock the run is being measured against is visible rather than implied.

One thing deliberately left alone: `_BoundSpoke` snapshots `start_time` at construction (`spoke.py:109`) to timestamp its bound trace file. That is a copy on a different object, only taken when `trace_prefix` is set, so a re-prep leaves it reading from the original construction.

Two tests in `TestPHPrepIsIdempotent`, both of which fail without the change: a first prep leaves the timer alone while a re-prep restarts it, and a second `ph_main()` still runs PH iterations after the timer is backdated past `time_limit`.

bknueven (approved):

PR #849, opened 2026-08-27
Debug synchronization

bknueven:


DLWoodruff:
# Review of "Isend/Irecv Backend for SPCommunicator"

**Review written by Claude (Fable 5.1), posted by David Woodruff.**
Document reviewed: [`doc/designs/isend_irecv_spcommunicator_backend.md` at 656f864d](https://github.com/Pyomo/mpi-sppy/blob/656f864d874a8aa349f75f66223c2a5bd36cb529/doc/designs/isend_irecv_spcommunicator_backend.md), the head of PR #849. Claims were checked against `spcommunicator.py`, `spwindow.py`, `hub.py`, `spoke.py`, `spin_the_wheel.py` and the extensions at that same commit. Line numbers below refer to the design doc.

## Summary

The direction is right and the doc is careful about the part that matters most: what must not change. Separating the layout exchange from `SPWindow` (which today does both jobs, confirmed), keeping write-ID and coherence logic above the transport, refusing a progress thread, and treating RMA as transitional are all good calls, and the phased rollout with differential testing is the right way to land it.

The protocol as written does not meet three of its own goals. The "one in flight, one pending" send rule does not bound receiver memory, because an `Isend` of an eager-sized message completes before the receiver has it (finding 1). The shutdown protocol counts completed sends, which for the same reason does not mean delivered messages (finding 3). And the flush lands in `free_windows`, which runs after `hub_finalize` reads the spokes' final bounds (finding 4). All three have small fixes. Findings are ranked by consequence.

## Findings

### 1. Eager sends defeat the memory bound and the backpressure (lines 232–252; Goals 5 and 9)

`MPI_Isend` completes locally as soon as the implementation has buffered the message, which for messages below the eager threshold (typically 8–64 KB depending on the MPI and the fabric) is immediate. Most of this traffic is eager: `SHUTDOWN`, the two bound fields and `BEST_OBJECTIVE_BOUNDS` are 1–2 doubles padded to 64 bytes, and a nonant vector of a thousand doubles is 8 KB.

So a hub publishing bounds every iteration to a spoke rank that is inside a long solve launches a new `Isend` every iteration. Each completes at once, the pending slot is never used, and the messages accumulate in the receiver's unexpected-message queue against a single posted `Irecv`. Receiver memory is then bounded by the publisher's rate, not by the design. Goal 5 holds only for rendezvous-sized fields.

**Recommendation.** Use `MPI_Issend` for the data stream. A synchronous send completes only when the receiver has matched it, which is exactly the backpressure the design assumes. With one in flight and one pending, the receiver side holds at most one matched message per stream, and "send completed" comes to mean "the receiver has it", which finding 3 also needs. The sender never waits on the request, so a send to a rank in a long solve simply stays in flight while the pending slot coalesces, which is the intended behavior. A credit or ack scheme would achieve the same with more code.

**Staleness trade-off of `Issend`.** Today a spoke that reads once per long iteration gets the hub's newest value at read time. Under synchronous sends with one in flight and one pending, a spoke waking from a long solve first matches the message that was in flight, which the hub launched when the previous one completed. The hub's newest snapshot sits in the pending slot and launches at the hub's next `progress` call, which can be up to one hub iteration later if the hub is in its solve. So a slow xhat or Lagrangian spoke can evaluate a value one hub iteration older than it would under RMA. That is an effect on bound quality, not wall clock. If the request-age metric planned in "Performance gates" shows it, the tunable is to use plain `Isend` for the tiny fields (`SHUTDOWN` and the bound fields, 64 bytes each, so unbounded queue growth is harmless in practice) and keep `Issend` for the large ones.

### 2. `progress` must drain, not repost once (lines 211–230)

Step 3 of the receive protocol reposts the `Irecv` after one completion. If more than one message is queued (always possible under finding 1, and still possible under `Issend` when the receiver's own progress calls are sparse), one test-and-repost per `progress` call surfaces one version per call, and a consumer walks through stale versions one call at a time. That contradicts "latest-value behavior". `progress` should loop test, copy to history, repost until the posted receive is incomplete, so the newest available snapshot is always the one offered. Say so in the doc; it is the kind of detail the unit tests with fake requests should pin.

### 3. Shutdown: drain by per-stream count, not by outstanding sends (lines 330–350)

Steps 3–6 reduce the outstanding send count to zero and then cancel unmatched receives. Under `Isend`, zero outstanding sends does not mean every message has been matched; eager messages can still sit in a receiver's unexpected queue. Cancelling the receive and freeing the communicator then leaves unmatched messages, which is what Goal 9 forbids (implementations mostly tolerate it, and the last snapshot is lost).

**Recommendation.** Each sender knows how many messages it launched on each stream (equivalently the last write ID it sent). Exchange those once in the closing collective (a fixed-size `Allgather` on the transport communicator, or one small message per edge) and have every receiver keep progressing until its received count matches for every stream. Only then cancel the still-posted receives, which are now certainly unmatched. This is exact, and under `Issend` it still earns its place as an assertion.

Also, `SHUTDOWN` is published exactly once (`Hub.send_terminate` is its only `put_send_buffer`), so the sentence at line 348 about coalescing replacing a pending shutdown snapshot describes a case that cannot occur. The publications that actually need the flush are the spokes' final `OBJECTIVE_*_BOUND` and `BEST_XHAT`, which `hub_finalize` consumes. Reword it around those.

### 4. Flush before `hub_finalize`, not in `free_windows` (lines 425–430, 330–346)

`WheelSpinner.run` at this commit does, in order: `main`, `send_terminate` on the hub, `finalize`, a `cylinder_comm` barrier, a `fullcomm` barrier, `hub_finalize` (which re-reads outer and inner bounds "to catch new values"), then `free_windows`. Phase 5 puts the collective close in `free_windows`. Under p2p, a spoke's last `send_bound` can still be in its pending slot after `finalize`, because nothing calls the transport's `progress` between that put and the barriers, and an MPI barrier drives the MPI engine but not the transport's launch-pending step. `hub_finalize` would read stale bounds.

Split the close: the flush half (steps 1–5) runs before `hub_finalize` as its own collective, the cancel-and-free half stays in `free_windows`. The RMA path has the same question and hides it because the barriers linearize the puts.

### 5. The candidate exchange should be typed, and the barrier can go (lines 279–309; Goal 6)

Step 3 "exchange candidate sets across `cylinder_comm`" reads as a Python-object allgather on every synchronized read, and the hub does several of those per spoke every iteration. With history bound K, allgather a fixed array of K IDs per rank (`-1` padded) and pick the greatest common ID locally. Same result, typed collective.

Today every synchronized read costs a `Barrier` plus two `Allreduce` calls (`get_receive_buffer` and `_write_ids_agree`). The barrier exists to make the RMA `Get` calls land close together; under p2p it has no purpose and the two reductions collapse into the one allgather. The doc says nothing about the barrier; state its removal as an expected win of the migration.

On liveness: the intersection is empty whenever some reader rank is two or more publications behind its peers at exchange time, and "wait for publisher quiescence" means "starve until termination", because the hub never quiesces earlier. In practice that needs ranks of one cylinder to stay skewed by two or more hub iterations, which is rare, but the phase 3 stress test should measure it, and the "rejected for empty intersection" counter planned in phase 3 is the right metric. The contingency design, if it turns out to matter, is a sender-side snapshot ring plus a pull for the agreed ID.

### 6. Size receive slots from the source's layout, not the receiver's (lines 211–230, 261–277)

On the unequal-rank path a peer rank's padded length for a local-sized field differs from the receiver's own, most visibly for `RECENT_XHATS` (versions times the source's block size). The equal-rank prototype never hits this because `_validate_recv_field` enforces equal sizes. "Receive state" should say the slot is allocated from the source rank's padded length in the layout catalog; it is the first thing to break in phase 4.

### 7. APH already runs MPI calls on a second thread (Goal 7; Application parity)

mpi4py initializes with `MPI_THREAD_MULTIPLE` by default, and the APH hub's `listener_util` runs `Allreduce` calls on a listener thread while the worker thread would be calling `transport.progress`. The transport state must be touched by the worker thread only, which holds if `progress` is called only from `put_send_buffer`, `get_receive_buffer` and the close. Write that constraint down. Also add APH to the parity runs: it is the one hub that reads with `synchronize=False` everywhere, so its read pattern differs from every other consumer listed.

### 8. Phase 1 will touch three tests, and the `window` alias is unnecessary (lines 389–395, 160–166)

"All existing tests to remain unchanged" will not hold. `test_flex_coherence_policy.py` stubs `sp.window` with `strata_buffer_layouts`, `set_write_ids` and `get`; `test_spwindow_multisource.py` and `test_spwindow_partial_get.py` construct `SPWindow` directly and rely on its allgather. Those three follow the layout extraction (the stub becomes a stub transport). Cheap, but say it.

Nothing outside `spcommunicator.py` reads `spcomm.window` (the `self.window` in `w_oscillation.py` is an unrelated moving window; `buffer_inspect.py` works on `FieldArray.window_array()`). The compatibility alias can be dropped from the plan.

### 9. Smaller points

- **Tags.** `Field.SHUTDOWN` is `-1000`, so the sorted-enumeration index is required, not a preference (MPI tags must be non-negative). Say why.
- **Regressing write IDs.** MPI guarantees non-overtaking per (source, tag, communicator), so a regression is a transport bug. Assert, do not warn.
- **Equal-rank global ranks.** With `_make_comms`, cylinder `c` rank `r` is global rank `r * n_spcomms + c`. Give that formula in "Equal-rank routes" so tests can check the catalog lookup against it.
- **Hub fan-out.** A hub rank publishes the same snapshot to every spoke rank at its position. One immutable copy shared across those edges is enough; one copy per edge is fine for correctness and only worth changing if hub memory ever matters.
- **`progress` at exit of `put_send_buffer`** (line 315) is redundant; publish launches the send itself.
- **Startup line.** Use `global_toc`, matching the existing "Using cylinders ... to solve a problem with ..." message.
- **Registration freeze.** Setup phase 3 before phase 4 is load-bearing: `xhat_feasibility_cut_extension.register_receive_fields` reads `fields_to_ranks` to find every emitting spoke. Keep the order and pin it in the test the doc already promises.

PR #856, opened 2026-09-09
New MPI RMA locking protocol

bknueven:
Replace mpi-sppy’s per-target RMA locking with a validated hybrid protocol:
  - Publishers use an exclusive self-lock, local stores, and Win.Sync.
  - Readers use short Lock_all / Get / Flush / Unlock_all epochs.
  - Remove pre-fetch cylinder barriers while retaining write-ID agreement.
  - Update mpi_one_sided_test.py to exercise the new protocol.
  - Add tests verifying exact RMA call ordering and absence of self-Put.

I would like further testing from @tvalenciaz before we merge.


bknueven:
Better, but not a complete solution

PR #864, opened 2026-09-16
Add a weekly Windows test workflow

DLWoodruff:
Adds two Windows workflows.

- `.github/workflows/test_windows.yml` runs a subset of the test suite on `windows-latest` with MS-MPI. It runs **weekly** (Mondays 07:00 UTC) plus `workflow_dispatch`, not on pull requests, so it adds nothing to PR turnaround. On failure it files or comments on a `weekly-windows-failure` issue, the same way `test_uc_weekly.yml` does.
- `.github/workflows/test_windows_quickstart.yml` runs on **every PR** and on pushes to main, and takes about 3 minutes. It installs the way `doc/src/quick_start.rst` says to for native Windows (MS-MPI runtime, conda-forge mpi4py, PowerShell) and runs that page's farmer commands verbatim: `--help`, the EF, and both PH runs. Each command is its own step, and every farmer step runs even when an earlier one failed.

## Weekly jobs

| Job | What it runs |
|---|---|
| Windows (serial and MS-MPI) | `test_ef_ph.py`, `test_config.py`, `test_sputils.py`, then `mpiexec -n 2 ... test_with_cylinders.py` |
| Windows straight_tests.py | `straight_tests.py` (spawns `mpiexec` itself) |
| Windows run_all.py *solver* (*part*) | `run_all.py` nouc for `xpress_persistent` and `xpress_direct`, first and second part |

MS-MPI 10.1 with mpi4py 4.1.2 from pip; `get_solver()` finds cplex. One step fails the job if no solver is found, so nothing quietly skips. The RMA windows in `spwindow.py` work under MS-MPI.

All six jobs pass: https://github.com/DLWoodruff/mpi-sppy-1/actions/runs/35032170873 (4 to 11.5 minutes each).

## The one-sided MPI test fails under MS-MPI

On `windows-latest` with MS-MPI 10.1.1 and conda-forge mpi4py 4.1.2, `mpiexec -n 2 mpi-sppy-one-sided-test` fails its first `assert(buff[-1] == 3.)`. In the same run, both farmer PH runs from the quick-start page pass: https://github.com/Pyomo/mpi-sppy/actions/runs/35379652255. The test reads rank 0's window while rank 0 sleeps without making any MPI calls. MS-MPI serves that read only when rank 0 calls MPI again, and mpi-sppy relies on the same pattern, so a busy spoke stalls the hub; measurements are in #869. `quick_start.rst` now says so, and the per-PR job leaves the test out, since it cannot pass under MS-MPI.

## Two fixes outside the workflows

These were needed to make the existing test harnesses work on Windows; both are no-ops on Linux.

- `examples/run_all.py`: `--branching-factors` arguments now use double quotes. `os.system` runs cmd.exe on Windows, which does not treat single quotes as quoting, so argparse saw `'3` and `3'`.
- `mpisppy/tests/straight_tests.py`: bash is located with `shutil.which`. A bare `"bash"` on Windows finds `System32\bash.exe`, the WSL launcher, and every command died before `mpiexec` ran.

Checked on Linux with gurobi: `run_all.py --second-part-only` reports "All OK"; `straight_tests.py` fails only the aircond command, which fails identically on a clean upstream/main.

## Notes for review

- The run_all first-part jobs build Pyomo's APPSI extensions, which netdes `--presolve` needs. That step runs under `pwsh` with `ilammy/msvc-dev-cmd`: setuptools does not find the runner's Visual Studio 18 on its own, and under `bash -el` the link step picks up Git's `/usr/bin/link.exe` instead of MSVC's.
- The weekly failure-issue job cannot run until this is on `main`; worth one manual `workflow_dispatch` run after merging.
- The per-PR quickstart job is a new check name; it is not a required check unless someone adds it.


bknueven:
Does this take some time to execute? I wouldn't mind having a regular Windows runner, even if it's a subset of what's here.

bknueven:
@DLWoodruff will add a little test which runs Windows on PR.

Issue #813, opened 2026-07-23
Decide the ruff rule set explicitly instead of relying on a version pin


DLWoodruff:
## Background

ruff 0.16.0 (released 2026-07-23) greatly expanded ruff's *default* rule set. Because the CI lint job installed whatever ruff was latest, every PR's lint check went red overnight with ~2,487 repo-wide findings in files the PRs never touched. #812 put out the fire by switching to the maintained `astral-sh/ruff-action@v3` pinned to ruff 0.15.21.

That pin is a stopgap. The repo should make a deliberate decision about which lint rules it actually wants, instead of implicitly inheriting whatever a ruff release ships as defaults. `.ruff.toml` currently only sets file exclusions — it selects no rules at all.

## Option A (small, recommended as the immediate follow-up): pin the *rules*, not just the version

Add the pre-0.16 default selection explicitly to `.ruff.toml`:

```toml
lint.select = ["E4", "E7", "E9", "F"]
```

This makes lint behavior version-independent: ruff upgrades stop being events, the CI version pin becomes belt-and-suspenders and can be bumped freely, and nothing in the codebase needs to change.

## Option B (larger, optional, can come after A): adopt some or all of the new 0.16 rules

Measured against ruff 0.16.0 (2026-07-23):

- 2,487 findings total under the new defaults.
- 1,170 are safely auto-fixable — but that diff alone touches **343 files / ~11,000 lines**.
- 984 more are "unsafe" fixes (need human review to confirm behavior is preserved).
- ~330 have no autofix (e.g. `TRY*`, `SIM*`, `B*` judgment calls).
- The biggest buckets are style: `UP032`/`UP031` (`.format()`/`%` → f-strings, ~400), `UP006` (typing generics), plus `TRY`, `FURB`, `PL*`, `RUF`, `EXE001`.

If adopted, this should be done rule-family by rule-family in dedicated mechanical PRs, **timed for when the open-PR queue is short** — a 343-file sweep conflicts with essentially every open branch. There is no urgency; Option A alone is a complete, stable end state if the new rules are judged not worth the churn.

## Acceptance

- [ ] Explicit `lint.select` in `.ruff.toml` (Option A)
- [ ] Decision recorded here on whether any 0.16 rule families are worth adopting (Option B), and follow-up PRs filed if so

PR #865, opened 2026-09-16
Select the ruff rules explicitly (#813 Option A)

DLWoodruff:
Closes the Option A half of #813.

`.ruff.toml` set only file exclusions, so the rule set was whatever the installed ruff happened to ship as its defaults. ruff 0.16 expanded those defaults and turned every PR red. This names the pre-0.16 defaults instead:

```toml
lint.select = ["E4", "E7", "E9", "F"]
```

pyflakes (F) plus the pycodestyle errors that are not pure formatting: E4 imports, E7 statement/comparison errors, E9 syntax and IO errors. No code changes.

## Verification

Run locally with ruff 0.15.21, the same version the CI lint job uses, and with ruff 0.16.7:

| | before | after |
|---|---|---|
| 0.15.21 enabled-rule list (`ruff check --show-settings`) | 61 lines | byte-identical |
| 0.15.21 `ruff check .` | clean | clean |
| 0.16.7 `ruff check .` | 2757 errors | clean |

So lint behavior today is unchanged, and it now survives the 0.16 default expansion. The version number in `.github/workflows/test_pr_and_main.yml` becomes belt-and-suspenders and can be raised on its own.

The other acceptance item in #813 -- whether any of the new 0.16 rule families are worth adopting (Option B) -- is a separate decision and is left to that issue.


bknueven:
I am fine merging this, but I would like to return to a blacklist for ruff rules once we adopt more of `ruff 0.16+`.

PR #868, opened 2026-09-18
adding power grid capacity expansion planning example (used for PMAPS…

tvalenciaz:
This is the example created for the PMAPS 2026 tutorial.
It is run in a jupyter notebook and consists in:
  - Solving a CEP problem in EF
  - Solving a CEP problem with PH sequentially (no MPI)
  - Solving a CEP problem with PH + cylinders, default rho (rel gap does not reach target)
  - Solving a CEP problem with PH + cylinders, sep-rho (rel gap reaches target)

This code was created by opencode using GPT 5.4, prompted by me. 
I have not reviewed or tested it extensively so it may contain uncaught errors. Use with caution.

Dependencies:
- highspy
To run the visualization:
- matplotlib
- scipy
- pandas
To run the notebook itself
- notebook
- ipykernel


bknueven:
We should consider adding this to the runall so its execution is checked

